PR #23, opened 2023-11-17
Created .gitignore for godot files

mango-gremlin:
## Motivation
Created a gitignore for godot files with the godot gitignore template.

## What was changed/added
added gitignore clauses for things we dont want to push with each commit.

I propose we try this out and then maybe add/remove things as we go, is everyone okay with this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+# Godot 4+ specific ignores
+.godot/
+
+# Godot-specific ignores
+.import/
+export.cfg
+export_presets.cfg
+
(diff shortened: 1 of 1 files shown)


Raining-Cloud (approved):
Yes a gitignore is essential, otherwise we would push temporary files. This is basically the godot template gitignore so let's just use this :)

mango-gremlin:
alrighty since its the recommended gitignore and no one seems to have a problem with it, i will merge this for now.

PR #26, opened 2023-11-22
Prototype without a Grid

mango-gremlin:
## Motivation
Grid was hard to implement, so we tried to implement the Bridge Prototype without a Grid.

## What was changed/added
@Kabooni @Raffael-Rack @marcel-iv @mango-gremlin Implemented a Prototype for the Bridge Building that doesnt feature a grid.

#### Features
- no grid
- no snapping
- no worries
- bridge mode and fox mode, fox can only move in fox mode
- button to toggle mode
- fox can move with a, d, space
- drag snake and / or deer, which becomes a platform for fox in fox mode.
- simple collision (doesnt work as intended)
- if fox touches water, it resets the position of the fox 
- if you pick up a snake, it displays the allowed placement zones for all snakes, but snapping does not work

## Testing
Should work for the aforementioned features but not for much else, will continue working on this 

see #27 for more details on what still needs to be implemented

#### Known Bugs
- input map forgets that it has assigned physics
- in the code for the fox (fox.gd), the left, right, jump buttons may not be assigned. Do this by going to Project Settings > Input Map > New Action and add left as A, right as D, jump as space
- in the input map for ui-accept, delete space as this can lead to the fox jumping activating the mode change button
- collision with draggable animals & each other & water does NOT work

## Additional Information
To import: copy folder somewhere else, zip it, then import in godot


aylinfi (approved):
great prototype! :0

ScatteredDrifter (approved):
You've shown and demonstrated your first prototype in the last high court meeting on 
**2023-11-23**. 
We've discussed your changes, implementations and considerations, and further also **denoted issues** that ought to be taken care of.

Considering that this is only a prototype to base further development, I am fine with merging even if it is introducing bugs ( which are self contained in its prototype instance so its fine )

Issue #29, opened 2023-11-22
fix snapping issues (snap to right place)


mango-gremlin:
## Motivation
The goal is to have the ability for animals to snap to each other in specific places, see #30 and #31 
For this to work, there need to be zones that correctly snap the dragged animal to the target animal. 

### Bug
So far there was a bug where the animal flew across the screen to (0,0) instead of snapping.

## Implementation
Find the bug and fix it. Possible causes:
- snappable area is child of draggable area
- wrong position caused by the fact that scene with target area is in world, but target area itself is not in world, but in scene with draggable animal -> position is (0,0) in that scene

mango-gremlin:
issue was fixed in #44

Issue #31, opened 2023-11-22
for each animal, save & check if you can place it in a zone


mango-gremlin:
## Motivation
why do we need this?

## Implementation
Animal you are currently dragging checks if you can place it in a zone
- create function can_place() that checks if the current zone is allowed for the animal one is dragging (called right when you release click, where it either snaps to the new place or the original one)
![image](https://github.com/mango-gremlin/arch-enemies/assets/104830903/ecd1e961-cd1f-41d3-9f89-54faec16e8c0)
call it here, and: edit the is_inside_dropable boolean, if the zone is not allowed for this animal -> set it to false

#### How to track which zones are which
-> assign each zone to a group (above, side, below)

Issue #30, opened 2023-11-22
for each animal, create zones where animals in general can be placed


mango-gremlin:
## Motivation
Until now we could only place animals at the left side of another one - now we need to extend this to all sides. 

## Implementation
A dropzone is implemented like this: 
- StaticBody2D
   - CollisionShape2D
   - ColorRect

The StaticBody2D is in the group "dropable" and "left_dropzone"/"top_dropzone"/"bottom_dropzone" and connected to the allowed_snake_area script.
At last, you have to add the new snapping position in the calculate_dropzone function of DragObject.gd, everything else is handled by the script.

PR #44, opened 2023-11-23
Fixed snapping issues and made snake snap to side of, not inside other snake

mango-gremlin:
## Motivation
closes issue #29 

Fixed the bug where the snake would fly off to point (0,0) when trying to snap to another snake. This is important since the snapping mechanic is integral to bridge building. Completion of Issues #30 and #31 is also dependent on the fixing of this bug.

## What was changed/added
To have the dragged animal snap to the target and not (0,0), we changed the position it snaps to from 'position' to 'global_position'. 

This made the snake snap inside the other snake. 
Fixed by:
When attempting to drop animal:
- save original position of snappable zone
- before snapping, switch x position of dropable zone to

dropable zone x =  
(drobable zone x) - ((width of dragged object) / 2) + (width of dropable zone / 2)

- when the object leaves the drobable zone, set the position of the zone to the original position it had before
- this is dependent on whether the zone was a side (right or left), or top or bottom

**currently this only works for the left zone of the snake. This should be taken into account when solving #30 and #31 & changed**

## Testing
Have tested it, it works when dragging a snake into another snake:
[Video of what works](https://github.com/mango-gremlin/arch-enemies/assets/104830903/a3f5543e-e46d-433d-9b0b-e584f86492be)

However so far it does not work when picking up the snake again, or if the snake that has been dropped into is picked up:
[Video of what doesnt work](https://github.com/mango-gremlin/arch-enemies/assets/104830903/e57fb99e-cfd4-43ac-bc7e-e943a2e544b4)

This has not been fixed as it is more part of #30 and #31.


Kabooni (approved):
Looks good to me, thank you for the effort!

Issue #45, opened 2023-11-23
prevent dropable zone from moving somewhere random when dragging animal


mango-gremlin:
## Motivation
in #29 we fixed the issues with snapping to the right place, the dropable zone of an animal. The implementation presented in #44 moved the dropable zone, which creates issues demonstrated in the PR.

## Implementation
Just take the coordinates of the new dropable zone, dont move the zone, and put the snake at the coordinates.

mango-gremlin:
bug fixed in #45

PR #46, opened 2023-11-23
fixed bug where draggable zones move randomly 

mango-gremlin:
## Motivation
closes issue #45 

Bug that needed to be fixed, should work as intended now.

## What was changed/added
Instead of moving the draggable zones, it saves the coordinates and just places the snake there -> no problems with the draggable zones moving about where they want.
Also changed the way the "zone_occupied" bool is updated, right after moving the snake to the  new place:
image

## Testing
No Problems when dragging something that had its draggable zone occupied:
[Screencast from 23.11.2023 21:42:46.webm](https://github.com/mango-gremlin/arch-enemies/assets/104830903/7c6b0ead-a0df-4c23-b1e8-24cafd535e38)


marcel-iv (approved):
bug is fixed, works well

Kabooni (approved):
Looks good to me

Issue #47, opened 2023-11-24
Snakes do not have collision anymore


mango-gremlin:
## Description
The snakes are supposed to be collision boxes for the fox to jump on. This does not work anymore.

**To Reproduce**
1. turn on fox mode
2. jump into snake

**Expected behavior**
Fox collides with snake

**Screenshots**
Video see #45 

## Solution Proposal
Add a staticbody node to the snake & a collision shape

mango-gremlin:
fixed in #48

PR #48, opened 2023-11-24
fixed no collision bug by changing type of snake parent node

mango-gremlin:
## Motivation
closes issue #47 

Snakes need to have collision for the fox platforming to work.

## What was changed/added
Changed type of snake parent node from Node2D to a StaticBody2D, and added a CollisionShape.

## Testing
Fox could jump on top of snakes.


Kabooni (approved):
Works great

marcel-iv (approved):
works as intended, seems to cause no further bugs

PR #53, opened 2023-11-25
Disabled dragging in fox mode

### LINKED ISSUE #28: no dragging in fox mode
## Motivation
why do we need this?

## Implementation
In the file global.gd, there is variable called drag_mode.
In DragObject -> before dragging check if drag_mode = true

Raffael-Rack:
## Motivation
closes #28 

Previously it was possible to drag the snakes while in fox mode. We want the player to switch between the bridge building and the bridge crossing, so that had to change.

## What was changed/added

I added a condition requiring drag mode to be true to the dragging function and the mouse-over highlight function.

## Testing

I tested the change in the Godot debug mode:
https://github.com/mango-gremlin/arch-enemies/assets/116198748/c092cb91-852b-49a5-bee3-49d3728003d8

## Additional Information
The changes are built onto the base project Prototype_Bridges_1.3


Kabooni (approved):
Looks good to me

PR #54, opened 2023-11-25
added grid snap functionality after releasing snake

### LINKED ISSUE #32: terrain snapping / grid snapping
## Implementation
After you let something go, check which position the object is in and snap it to the closest of a grid of pre-defined coordinates.

Ex you are in (12,13), it snaps to (10, 10) -> modulo
Use a variable

marcel-iv:
## Motivation
closes #32 

## What was changed/added
Dragged snake now snaps to nearest position within grid after releasing the click button, before considering if it is within a placeable zone or not. Currently a grid square size of 10x10 pixels is assumed, this can be changed via the ``grid_size`` variable. Getting the nearest grid square is done using the ``round_to_nearest(a:float, b:float)`` function. 

Also, initial placements of snakes has been adjusted to the 10x10 grid.

## Testing
The effect is barely noticable, as it occurs in the same frame as the snapping to droppable/return to original position. Using print statements, however, confirms the functionality of the code (not included in commit, but printing ``self.position`` before and after the snap can be tested).

image

## Discussion
Adds the helper function ``round_to_nearest(a:float, b:float)``, which may be helpful elsewhere. It may also be beneficial to put all helper functions in an external file.

### DIFF
--- a/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/DragObject.gd
+++ b/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/DragObject.gd
@@ -9,10 +9,10 @@ var initial_pos : Vector2
 var is_dragging = false
 var inside_object = false
 var dropzone_left_occupied = false
+var grid_size : float = 10.0 # size of a square in grid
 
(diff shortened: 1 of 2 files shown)


aylinfi (changes requested):
Works great! Maybe put the helper function round_to_nearest into global.gd for global use.

aylinfi (approved):
Changes will be done in a new issue, so we can approve this now.

Issue #34, opened 2023-11-22
hazards that kill fox


mango-gremlin:
## Implementation
When fox touches hazard, reset position of fox 

In the file fox.gd:
Make collisionshape2D, when entered, call function reset_fox() -> resets fox to original position

PR #55, opened 2023-11-25
Resolved Issue #34 Simple Hazard

koennixer:
## Motivation
closes issue #34

## What was changed/added
Added an additional CollisionShape2D to the main area, that resolves like the water below.
The CollisionShape2D also has an associated prototype sprite. This is subject to change.

## Testing
The Shape works in testing with the current main branch.

## Additional Information
This is not a fully implemented version of the concept of Hazards. Ideally there would be a generic Hazard that can be adapted into this. Secondly the should be some sound to alert the player to them having run into the Hazard. Lastly the animals should not be able to be placed into the Hazard, but since the animal placement is subject to change adding this now is not sensible.


Kabooni (commented):
It works great, but it would be better if you seperate the hazard into its own scene. It will be easier for us later, when we have multiple ones

mango-gremlin (changes requested):
i agree, maybe make a new scene for the hazards so we can later add them more easily

however i love the hazard sprite we should def keep it for the prototype :heart_eyes:

mango-gremlin (approved):
looks good and works, maybe include a video next time so we can see the changes :)

PR #70, opened 2023-11-28
Fixed: dragged object should be on top level

### LINKED ISSUE #67: Animal that you are currently dragging should be on top
## Description
When dragging, the animal that you are holding isn't always in the front and sometimes behind other objects.

**To Reproduce**
1. Drag one snake over other snake

**Expected behavior**
Animal is on top so it is visible

## Solution Proposal
Set the top_level property to true, so that the animal will be on top of everthing else.

aylinfi:
## Motivation
closes #67

When picking up the object (e.g. a snake) it might be displayed behind other objects. To make the currently dragged object more visible and stand out more it should always be displayed on top of everything else.

## What was changed/added
In DragObject.gd the top_level of the dragged object will be set to true if the action is just pressed and it will be set to false if the action is just released.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/44772f8c-ab2a-4366-bff7-158e2c134668


Kabooni (approved):
Works great, very simple solution

Issue #69, opened 2023-11-28
Change segmented hitbox based approach to grid-snap based approach


marcel-iv:
## Motivation
Our current implementation with segmented hitboxes is hard to implement, and struggles with placement of larger objects, especially in larger objects. I came up with another approach, that should resolve the issue entirely, and be able to allow creation of new animals a bit more easily. However, this would also change a few other issues, so this issue will not be a priority for now.

## Implementation
When dropping an object, snap to nearest grid square. Then, check if the dragged body is currently within an allowed drop zone, and *not* within a non-allowed drop zone (such as a snake's head). If yes, then leave the object there, else, snap back to the original position. Draw a grid when dragging an object, and the grid size of a square is based on the height of the snake (so that a snake is 1x5 squares long, for example). Also create new drop zones for the existing animal, as well as the forbidden drop zones.

PR #71, opened 2023-11-28
implemented squirrel

### LINKED ISSUE #65: implement squirrel
## Motivation
We want squirrel.
So the prototype level is playable (only snakes doesnt work because of the head limitation)

## Implementation
Implement like the snake.

mango-gremlin:
## Motivation
closes #65 

We need a squirrel to make the prototype level completable.

## What was changed/added
Added a squirrel with the same structure as the snake
- Staticbody with a collision shape
- an area2d with a collision shape
- more staticbodies for the dropzones
- sprite

Connected the script for Dragging (DragObject.gd) to the squirrel and the dropzone script (allowed_snake_area.gd) to the dropzones. The squirrel is implemented in exactly the same way as the snake.

Also changed the calculate_dropzone function to calculate the dropzone dynamically from the size of the currently dragged animal and dropzone.

## Testing
[Screencast from 28.11.2023 14:28:44.webm](https://github.com/mango-gremlin/arch-enemies/assets/104830903/b850fc5f-141c-4d44-97c1-049c97d64ba9)

Squirrel works in the same way as snake, can be placed on snake, snake can be placed on squirrel.

## Additional Information
this may become redundant in if #69 is finished


Kabooni (approved):
Lgtm

PR #73, opened 2023-11-28
Fixed Dropbox collision bug

### LINKED ISSUE #49: Fox collides with dropable zones
## Description
Since the Dropable zones of animals are staticbodies with collisionshapes, the fox collides with them too.

**To Reproduce**
1. Turn on fox mode
2. jump on area left to snake
3. float

**Expected behavior**
Fox collisions ignore dropable zones, fox cant stand on them.

## Solution Proposal
Using collision layers, put the dropable zones in a different collision layer than the fox.

koennixer:
## Motivation
closes #49
The Dropbox collision bug was a major bug prohibiting proper jump/run 

## What was changed
To fix the Bug all the dropboxes are now on Physics layer two. The main snake collision box now is on both Physics layer one and two. 

In the main document the Physics layer one and two have been renamed to showcases their purpose.

## Testing

fox_is_not_bugged

## Additional Notes
The behavior and drag and drop feels very janky. It behaves the same as before as far as I can tell and test, but I cannot guarantee that the behavior hasn't changed in some edge cases.

Resolves https://github.com/mango-gremlin/arch-enemies/issues/49

### DIFF
--- a/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/project.godot
+++ b/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/project.godot
@@ -56,6 +56,11 @@ jump={
 ]
 }
 
+[layer_names]
+
+2d_physics/layer_1="collision"
+2d_physics/layer_2="droppable"
+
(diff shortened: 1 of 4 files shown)


mango-gremlin (approved):
Lgtm, nice!

I think the drag and drop was janky from the beginning, as far as I could tell it isn't jankier than before. (this may be fixed by #69)

PR #74, opened 2023-11-29
Moved the round_to_nearest function to global

### LINKED ISSUE #63: Move helper functions to global file
## Motivation
As discussed in #54, the helper function should be moved to access it more easily in different parts of the code. 

## Implementation
Move the function to the global.gd file and use the ``load`` function to access the function in the code. Make sure everything works properly. 

## Discussion
If you know of any other helper function that should be moved, let me know or commit to the branch directly. In the future, all helper functions should be collected there.

Raffael-Rack:
## Motivation
closes #63 

Moving the round_to_nearest function to a global file makes the code more legible and allows us to more easily call it from different scripts.

## What was changed/added

I moved the round_to_nearest to the global script and updated the function calls to call it from global.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116198748/7ea887d4-cb4c-4113-bf3a-d1fb025f5192

The game still works.

## Additional Information
If any other functions should be moved mention it.


marcel-iv (approved):
Works just as before, good refactor. I'm not aware of any other functions on main that would benefit from a similar refactor, although the overworld group might have some, and future such functions should directly be written in global.gd

mango-gremlin (approved):
looks good. I cant think of any functions rn that should be in global as well but we should keep that in mind when we make new functions.

PR #75, opened 2023-11-29
Issue69/grid based approach

### LINKED ISSUE #69: Change segmented hitbox based approach to grid-snap based approach

marcel-iv:
## Motivation
closes #69, allows much easier implementation of animals in the future, and the easier handling of #31.

## What was changed/added
Instead of snapping to pre-defined collision zones, the animals now only check if their current position is valid (rudimentary for now, specifics will be done in #31). This makes it much easier to create new animals, and makes placement a lot more intuitive. Also, groundwork for #31 was done. The code still contains print statements, as I believe it will be useful in handling some rather annoying bugs.

Also note that unused code for the previous code was removed. If it becomes relevant again, it can easily be found by referencing earlier commits. 

For easier testing, a grid was drawn in the background, although this is done via a simple 500x500 sprite. The associated code can be used in the future, but the sprite should be replaced with a bigger one.

## Testing
[Demonstrate the code is solid. ](https://github.com/mango-gremlin/arch-enemies/assets/104799849/d4969e35-065f-4043-b782-cb632312e25e)

## Bugs
Three notable bugs have been identified, and should be handled *soon*. Bug reports will be created shortly.


Kabooni (approved):
Very nice solution, code looks good

PR #78, opened 2023-11-29
check validity of placement of animals to each other

### LINKED ISSUE #31: for each animal, save & check if you can place it in a zone

marcel-iv:
**IMPORTANT: This is based on PR #75, so first approve and merge #75 before this one**

## Motivation
closes #31 

## What was changed/added
Checks if the drop zones are considered valid. Per default, animals can only be on top of each other, but squirrels can stick to sides, and spiders to everything, and everything can stick to spiders.

I also wrote a small helper function to identiy animals based on their node, as I didn't find an appropriate integrated function in Godot. If such a function exists, feel free to let me know, so I can correct it.

## Testing
Here the code for squirrel and snake is demonstrated.
https://github.com/mango-gremlin/arch-enemies/assets/104799849/5c3688fa-5db9-4218-9c1c-a373163e2bf1

Here, the validity for the spider's functionality is confirmed. The squirrel here is tagged as a spider, so it behaves exactly like a spider would. 
https://github.com/mango-gremlin/arch-enemies/assets/104799849/5b308f09-ee8b-4244-87b9-354ffbb3a44c

## Additional Information
This code allows us to easily create a functional spider, which we should do in a separate issue to demonstrate more of our game. To do this, simply copy the squirrel, adjust sprite and size, and tag the root node as *spider*.


Kabooni (approved):
Looks good to me, it's very nice that you already implemented the rules for the spider!

PR #79, opened 2023-11-30
added multible drop zones to the shore and a new Group: right_dropzone

### LINKED ISSUE #64: Add drop zones to shores
## Motivation
To properly play the game, we need drop zones not only for animals, but also for (allowed) terrain.

## Implementation
Add collision2D shapes to the shores, just like the dropable zones for the snakes. See #52 
Add shapes to group ``dropable``

trashsolid123:
## Motivation
closes #64

now Bridges can start at the Shore

## What was changed/added
added 5 drop zones to the Shore, the new drop zones are in the new Group right_dropzone which calculates the correct possition. 

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/44133213/b8b99439-5188-40ca-ac79-e912aa10f87a


marcel-iv (changes requested):
The code works well, however, it is based on a version of main that is out of date, and still uses the old approach. Please merge main into your branch and implement it in the new approach. To do that, give the dropzone the group *dropable*, and *drop_dropzone*/*side_dropzone* for the different collision shapes. If you run into any issues due to my change in approach, let me know

PR #85, opened 2023-11-30
Made the grid better visible

### LINKED ISSUE #62: Draw a grid in the background when something is being dragged
## Motivation
As described in #54 , the snappable grid is not very noticeable. To make it clear that there is a grid, draw a grid in the background.

## Implementation
Replace the existing Grid sprite with a better grid (should cover the viewport and be 10x10 pixels for each grid square), but keep the current name, as the grid already hides itself when nothing is being dragged.

koennixer:
## Motivation
closes #62 

The grid was barely visible, this should improve that.

## What was changed/added
Added additional tiles for the background.
Expanded the background and put it on it's own z layer to avoid collision with foreground objects.

## Testing

grid

## Additional Information
This is not implemented as requested. The request would requires making the grid individually flexible in color to adjust to the background, this is tedious and means the grid would have to change every time something about the background is changed. If there was a different intention, I can adjust the changes.


marcel-iv (changes requested):
Adding a background is very nice, but from what I can tell, the grid remains unchanged, which was why #62 was opened in the first place.

Since the issue might be a bit misleading, let me rephrase the task here: You should simply make the existing grid larger, so that it covers our level/viewport. This could simply be done by using a larger sprite for the grid, but other, more dynamic options are certainly possible, too. But for  a prototype, simply using a larger grid by stitching the existing sprite together would probably be enough.

Also, adding the grid to a separate z-layer was a great idea, if the grid is just a bit larger, we can merge to main.

koennixer:
I have attempted a fix. I have simple place an additional grid using the same texture. This should fix the problem of making the grid bigger.

PR #86, opened 2023-12-01
Snake now snaps into dropzone

### LINKED ISSUE #76: Snapping allows placement of invalid positions
## Description

**To Reproduce**
1. drag one snake over the drop zones of another animal
2. drop the snake, while it is still touching the drop zone, but with the dragged snake being on the outmost part of a grid square
3. snake will snap to outer grid square (outside of the drop zone), but will remain there and not snap back

**Expected behavior**
Snake snaps to outer grid square and resets to original position, as the snapped position is not valid.

**Screenshots**
https://github.com/mango-gremlin/arch-enemies/assets/104799849/6de7d2c6-f273-405a-911f-02209e5831e5

## Solution Proposal
The problem arises, because the snap to grid happens in the _process function, so the variables is_inside_dragable and is_inside_forbidden are not updated, as the dragged animal hasn't moved. Everything occurs in the same frame, so no update is possible. To fix this, maybe use signals emitted by events to handle the snap to grid, so it occurs in different frames (or rather, before _process is automatically called)

mango-gremlin:
## Motivation
closes #76
Snake did not snap into dropzone if it was let go only touching the edge of the dropzone.

## What was changed/added
Created function snap_into_dropzone(x,y) which in theory calculates how to change position of currently dragged animal so it snaps into dropzone if it only barely touches it.
This works most of the time now if you don't go out of your way to place it on the veeery edge.

**Implementation**
- generate a new area2d with a collisionshape which is at the exact position calculated by round_to_nearest, and the same size as the object we're dragging
- if this shape is not overlapping with the dropzone, we adjust the round_to_nearest input

1. if the object we're dragging is to the side of the dropzone -> so x doesnt need to be changed, so only change y
2. same for the other way around -> change x but not y
3. change both to the position that places it inside the dropzone
-> if y above dropzone, round up (so object is lower)
-> if y below dropzone, round down (so object is higher)
(same for x)

- delete the area2d at the end to avoid shenanigans. & avoid clutter

We also changed these: (if these were solved in another pr, overwrite them, as they are provisional changes we had to make to test the code)
- Adjusted parameter names in round to nearest (to make clearer what they are)
- adjusted background grid so it covers whole level
- changed squirrel hitbox and size so it matches grid

## Testing
https://github.com/mango-gremlin/arch-enemies/assets/104830903/99feadf7-1899-4ab4-b3a6-91831bd7f9b4

As you can see this doesnt work reliably when placing from the top or edge. But it does work well when placing from the side.
At some point we did get it to work for the edges, but implementing the functionality for the side broke it.
This is something that needs more time -> after prototype presentation?

**There is also the problem where snakes can be placed on top of snakes if none of the dragged snake touches the head zone of the stationary snake -> this was also the case before.**

## Additional Information
Implementing this perfectly entails adding about 3934502314 different conditions, which would be extremely time-consuming, so maybe this is something for later. It works well enough most of the time.

### DIFF
--- a/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/DragObject.gd
+++ b/testing_area/prototype_testing/MarcelRaffaelKatjaGlenn_BridgeNoGridPrototype1/Prototype_Bridges_1.3/DragObject.gd
@@ -75,9 +75,11 @@ func _process(_delta):
 			new_position_x = Global.round_to_nearest(self.position.x, grid_size)
 			new_position_y = Global.round_to_nearest(self.position.y, grid_size)
 			
+			var new_position : Vector2
+			
 			if self.dropzone != null:
 				
-				var new_position = snap_into_dropzone(new_position_x, new_position_y)
+				new_position = snap_into_dropzone(new_position_x, new_position_y)
 				
 			# 1. nothing is being currently dragged
 			self.is_dragging = false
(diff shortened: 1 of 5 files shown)


mango-gremlin:
Last night we refactored our code into a new function: snap_into_dropzone
but the value calculated by the function was never used. This is why snapping didnt work well at all.
The new value is now used in the new position:
- snap_into_dropzone calculates the vector new_position
- the snapping animation now also snaps to the new_position and not the old one, as previously

Snapping should now work for:
- the corners of dropzones
- the sides of dropzones
- top still doesnt work all of the time.

marcel-iv:
Will close this PR, as the bug was alreay fixed in #97

PR #88, opened 2023-12-01
added drop zones to shore

### LINKED ISSUE #64: Add drop zones to shores

trashsolid123:
## Motivation
closes #64

## What was changed/added
added drop zones to the Shore with new Group "shore_dropzone"

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/44133213/ef36e82a-defa-454e-a7d5-ffd333f1b86f


aylinfi:
Adding dropzones on the right side of the shore would be good too.

Issue #92, opened 2023-12-01
Overworld : save / load inventory with items of type "Item"


ScatteredDrifter:
## Motivation

The current version of the save and loading mechanism is not capable of **saving** and **loading** an inventory based on entries of type "Item".

I've commented out the capability to avoid errors as its not necessary yet.  

## Consideration:

To resolve this issue we have to:
1. **save** an item with its according properties in our save-file 
2. **load** properties of item and **generate a new instance** that is representing this item accordingly.

ScatteredDrifter:
resolved in #99

Issue #90, opened 2023-12-01
overworld : refinement 


ScatteredDrifter:
## Motivation
We need to incorporate some things that are required for correct functioning 

This is a convolution of smaller aspects, tasks that I wanted to collect in order to resolve them one by one. 

## Aspects: 

- [x] add bridge-ID  to interaction-area decendants that are of type Bridge! 
- [x] make those bridge-IDs queriable for the player to switch to
- [x] create inventory to contain items instead of strings 
- [x] obtaining item is denoted by pattern matching against enum Item

Issue #59, opened 2023-11-26
replace string matching for interaction-decisions 


ScatteredDrifter:
## Motivation

Currently the players actions on how to interact with an area that was detected is handled by **matching against Strings**.
This might work for a prototype yet could create issues the more we add later on. 

## Consideration

Instead it might be useful to replace this structure with **pattern matching against an enum / case class**.

Issue #56, opened 2023-11-26
save state when leaving overworld 


ScatteredDrifter:
## Motivation

As of now our player character is reset and spawns at (0,0) whenever we leave the scene ( i.e. entering game mode, opening menu etc. ). 

However we ought to save the state of our player in order to have a more pleasant experience and to allow for saving states between different runs. 

## Implementation

We do require an instance that is initialized whenever the game is started that will hold ( and load ) the following: 
- choosen name ( might not be necessary but for testing  
- current inventory 
- position in overworld 
- settings ( volume etcetc ) 
- player stats like health, etc 
- ( _for the card game:_ current deck, special modifiers ) 
- progress of game ( "reached X island alread") 
- world state ( progress as mentioned above, interaction levels with npcs, availability of items scattered around ) 

There are multiple considerations to achieve this: 

### Consideration: Player-instance as backbone 

We could store those in the player node and propagate the information to other nodes as needed. 
as of now a method to "save the state" is implemented which could then save everything on change.

### Consideration: main-scene as backbone:

Considering that these information are necessary **not just the player** we might just incorporate this state retrieval  
into our mainscene and **propagate** those into the designated nodes, or have them draw the information whenever required. 

### Where to Save? 

I think simply creating a YAML / Json 
to store our current state is the easiest solution.
Whenever a "save state" is initiated we can then take the current snapshot and store it somewhere

That way people could modify it as needed, however we should not worry about as of now.

ScatteredDrifter:
Referenced  in #89

PR #93, opened 2023-12-01
Prototype/overworld interaction refactoring

ScatteredDrifter:
## Motivation
This PR modifies several parts of our overworld system. 

Resolves the following issue:
#59 
#90 
#56 -> although this has been added in "prototype/overworld" and de34c4b

---

## Disclaimer:
This is an internal pull request:

Its not merged into our main-repository as its not supplying a functional and completely working prototype. 
It runs and works mostly, yet we have some minor issues like _refactoring_ _bad structure_ or similar that needs to be worked on.

## Changes: 
This branch modified and updated the "player-interaction" in several ways:
- interactions are now denoted by a **Type** ( an enum)
- the player interaction is decided based on the given **interaction-type** and handed to the given handler 
- the inventory now contains entries of type "Item" 

Because the player-interaction was changed I also changed the inventory and its contained items:
- the player-inventory is now an "Array[Item]" of Items. 
- updated adding and deleting an item accordingly
- updated **UI** to correctly query names of items in the inventory 

However because the saving system was implemented simultaneously, we cannot save the inventory as of now. Further details are given in #92

## Testing

Runs well, been tested in different scenarios. 
**what does not work**: 
-> saving inventory, see #92

### DIFF
--- a/testing_area/prototype_testing/overworld/entities/interaction_overworld/bridge_interaction.gd
+++ b/testing_area/prototype_testing/overworld/entities/interaction_overworld/bridge_interaction.gd
@@ -1,0 +1,15 @@
+extends Node2D
+
+## base properties set for bridge interaction
+
+@export var bridge_id:int
+@export var difficulty:int
+@export var description:String = "reserved for preview path of sprite"
+
+func _ready():
+	var interactionspot_object = get_node("interactionspot")
+	
+	# updating bridge id accordingly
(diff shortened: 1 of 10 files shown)


ScatteredDrifter:
With the latest commit d5aed8a the issue of positioning the save location to another place was updated. 

It is not contained within its own script and used as a class that is instantiated upon starting the game  ( from the maingame scene )

PR #84, opened 2023-11-30
Added a spider

marcel-iv:
## Motivation
closes #81

## What was changed/added
Added a spider as a scene, and added an instance of it to the main scene

## Testing
https://github.com/mango-gremlin/arch-enemies/assets/104799849/9d11b5ad-c18d-4892-b30c-6f2fcc20d100

PR #101, opened 2023-12-03
Merge the prototype PRs

### LINKED ISSUE #62: Draw a grid in the background when something is being dragged

### LINKED ISSUE #76: Snapping allows placement of invalid positions

### LINKED ISSUE #77: squirrel snaps to wrong grid
## Description
Some sprites with certain sizes (such as the squirrel's 2x1 size) is affected differently by the grid snapping, as their middle point/origin is on the edge between grid squares, and not in the centre.

**To Reproduce**
1. Drag a squirrel on a snake
2. Watch as it hovers menacingly 5 pixels above the snake

**Expected behavior**
The squirrel should snap to the snake's back, just as the other snake

**Screenshots**

https://github.com/mango-gremlin/arch-enemies/assets/104799849/dd13edda-3eb1-4136-98e6-4aa106797435

## Solution Proposal
Update the snapping logic for the squirrel in some way, maybe a variant of the function only for the squirrel (you can check animal type by groups). Also simply adjusting the position of the squirrel within its own scene might just be enough to fix it.

### LINKED ISSUE #81: Add a spider
## Motivation
Add a spider, as all coding has been done for it, and it expands our prototype to be a lot more interesting

## Implementation
Copy squirrel, change size of squirrel to be 1x1, update sprite to spider, update group of root node to *spider*

### LINKED ISSUE #82: Make game solvable
## Motivation
We need to be able to win our game to demonstrate it, as currently it is impossible to do so.

## Implementation
Add more animals and adjust fox's jumping height so the game become actually solvable. If you place animals, please place them appropriate for our grid size.

### LINKED ISSUE #96: Bridges: Continous Gridsnapping
## Motivation

Currently we are calculating what point in the grid an animal snaps to in the same step where we determine whether its location is valid, when dropping it. This has led to problems where the animal would register that it was in a valid place but then snap to a position outside the drop zone.
By making the snapping occur while dragging the animal the player will be able to see the final placement of the animal before letting go of it and we avoid issues that stem from doing all of the calculations at once.

## Implementation

The current code for dragging sets the position of the animal to be equal to that of the mouse, up to an offset based on where the animal was picked up. 
Instead I want to create a new variable that represents the position of the mouse rounded to our grid size. The animal will then be moved to that position.

marcel-iv:
## Motivation
closes #96 
closes #82 
closes #77 
closes #81 
closes #76 
closes #62

## What was changed/added
Merged the following PRs into a separate branch for testing, so that main doesn't break. These PRs include: #97, #95, #94, #85, #84

Note: Issue #64 still had some bugs in combination with other PRs, so it is not merged into main for now. @trashsolid123 will keep us updated on the issue. 

## Testing
The Spider isn't show here, @Raffael-Rack will update #82 so that the spider is also included in the final presentation.
https://github.com/mango-gremlin/arch-enemies/assets/104799849/784876d4-b840-4547-b075-cbd7b3f3432a


Kabooni (approved):
Nothing breaks, lgtm

PR #105, opened 2023-12-03
Added a button that spawns spiders.

### LINKED ISSUE #102: Adding a button to create spiders
## Motivation

Now that there are spiders, there needs to be a button to spawn them.

## Implementation

In the same way, as the other buttons.

Raffael-Rack:
## Motivation
closes #102 

We added spiders to the game, so we need a button to spawn them so they can be used to build the bridge.

## What was changed/added

I added a button that is connected to a script that instantiates a spider when the button is pressed.
I also changed the size and position of the button, because we want to make the level smaller and sorted all of the control element nodes to be next to each other in the scene tree.

## Testing

Conjuring a spider:

https://github.com/mango-gremlin/arch-enemies/assets/116198748/6df52cf2-a9ac-46eb-8003-2818861bfb02


Kabooni (approved):
lgtm

PR #110, opened 2023-12-04
adjusted snake hitboxes to fit grid

### LINKED ISSUE #104: Align snake to grid
## Motivation
Currently the snake and its collision zones are not aligned to the grid, which could maybe be the cause of bugs. Also, it looks a bit weird currently. 

## Implementation
In #86, this was already fixed, however, we didn't merge the PR, as the core bug became a non-issue. To solve this issue, simply find the changes in the snake scene and overwrite the current snake and check if that causes any bugs (it shouldn't, but just to be safe)

marcel-iv:
## Motivation
closes #104 

## What was changed/added
Slightly adjusted snake hitboxes so they (visually) fit the grid. Took most of the work from #86, adjusted it only slighty 

## Testing
As seen below, the snake and its hitboxes fits the grid 
https://github.com/mango-gremlin/arch-enemies/assets/104799849/2a0208a8-58cd-4027-928b-bbb3626108b6

## Additional Information
As you can see below, the hitboxes still don't look pretty internall, especially considering the centering of the collision zones. After the prototypes have been merged together, we might want to fix that. 
image
Also, the red rectangle is now brown thanks to the background, it doesn't look good.


Kabooni (approved):
Code looks good

PR #117, opened 2023-12-04
Feature/merge prototypes

ScatteredDrifter:
## Motivation

We established the base structure for the _now combined_ project.
For that reason we merged both the overworld and bridge prototype to be contained within the **same project**.

## What was changed/added

In pair programming @mango-gremlin  and @ScatteredDrifter merged both the prototypes for the **overworld** and the first bridge level.

> While merging minor issues occurred ( linking issues, global variables not available ) which were fixed in below's code. 

--> We adapted the folder structure from #2, yet slightly adapted to separate each large compartment ( bridge, combat, overworld ).
This will help to minimize overlapping code spaces and also to organize it all a little better. 

> The main-scene (main menu) is located at the root of this document.
> it is also set as default scene to execute now

## Testing

Tested by both of us, the old bugs remain, yet are tracked in this repo already.


mango-gremlin (approved):
all the stuff that worked before still works, bridge bugs remain but we were aware of those.

Issue #139, opened 2023-12-11
overworld | change inventory system


ScatteredDrifter:
## Motivation
Instead of collecting items individually, we should only count **how many** we have collected of a given type.

This makes it easier to query this inventory, requires less space and we dont need the complexity of objects ( or their possibilities ) 

## Implementation
This change is mostly done in **Player.gd** 

change inventory to be a list / dictionary that only contains the amount of items owned ( of each type we have ) 

**Important**:
This is a fundamental change and thus requires **updating** methods that interact / change the inventory. Specifically that is: 

-> add to inventory 
-> remove item 
-> search item

Issue #136, opened 2023-12-11
overworld | values contained on "bridge_interaction" 


ScatteredDrifter:
## Motivation

The "bridge_interaction"-spot is an area where the player can enter or reach out to a bridge level.

-> Here a bridge level is denoted by a given "id" which links said interaction with the corresponding level.

**Question** do we require further Information to be stored there? 

Some ideas I've come up with: 
**reward upon completion** -> could be either an animal, item or similar ? 

## Implementation

As of now in`/overworld/entities/interaction_overworld/bridge_interaction.gd` the base structure for itneracting with a bridge-interactionspot was set.

We should implement further information from there

ScatteredDrifter:
Considering that I'm restructuring the `interactionarea` completely once again in the feature branch resolving #134 
I came up with the following structure: 

Upon **interaction with a bridge-area** one gets the following structure: 

Object of **InteractionValue** containing: 
- InteractionType.BRIDGE --> to handle interaction accordingly
- Dictionary Of Values: 
   - "bridge_edge" maps to --> Object of **BridgeEdge** to represent the bridge in graph 
   - "text" maps to --> String containing information about level ( like s small description ) 
   - "issolved" --> Boolean denoting whether bridge-level was solved already or not

With those information the player could then continue to interact with this area, as in entering the level or not.

ScatteredDrifter:
This is now also provided and documented in the wiki: 
https://github.com/mango-gremlin/arch-enemies/wiki/04-Overworld-Navigation

ScatteredDrifter:
@Kabooni @mango-gremlin Could you iterate on those information? 
Do we require information or is it fine like that? 

The proposed state in my comment is currently part of **main** so its already available and you could check whether thats fine for you.

mango-gremlin:
so since we are not doing the level reward at the moment, just saving the bridge-id is fine for now, this works already, so I would say this is okay for now

ScatteredDrifter:
Oki, I will close it then.

Issue #133, opened 2023-12-10
overworld | gamedesign | set requirements for obtaining items


ScatteredDrifter:
## Motivation

It might be useful to set certain requirements to prevent the player to **obtain** an item from certain spots if **specific tasks** were not complete or certain thresholds where not met yet. 

For example one could prevent the player to obtain item "Honey" by requiring them to reach a given level of friendship with faction X or similar. 

Those limitations should be considered to **create motivation** and some **difficulty** to prevent grinding items. 

## Implementation

The **interaction area** node which can be instantiated and filled with its own values **exports** a variable that denotes whether a limitation is set or not.

ScatteredDrifter:
Considering that Items are only used once and don't have any requirement, like **factions** to be friends with or similar, we can omit this feature and close this issue.

Issue #119, opened 2023-12-07
Tasklist Assets Overworld


mango-gremlin:
### Deadline
9.1.

## To Do
- [x] insert checklist

added on 2023-12-11:
### Basic Tilemap Elements ((2.5D))
   - [x] Grass, Steine, Flowers, Dirt, 
   - [x] Übergänge zwischen unterschiedlichen environments
   - [x] Bäume, Sträucher, Schilf 
   - [x] Wasser, Fluss-Elemente -> Steine im Wasser
   - [x] Klippe oder ähnliche Begrenzung der Welt 

### Interactable Areas/Bridges/Items
   - [x] Tier-Unterschlupf ( entweder Häuser oder tierische Unterkünfte)
   - [x] Brücken-Ansatz
   - [x] Gebaute Brücke

For everything above: see

So far not 100% decided:
   - Big Rocks (not gonna be done)
   - [x] #257 

### Player Animation
- [x] Animate Fox for Overworld #274

Player-Sprite ( 4-way movement ) -> fox will be animated in x direction by bridges UI team, and when that's done, I will add the y direction so the fox can move in all 4 directions needed for the overworld navigation

### UI
- [x] **Basic Menu** 
   - [x] Dialog-Feld 
   - [x] Legenden für Einstellungen ( Sound )

## Additional Information
For the comprehensive List of Assets see #10

ScatteredDrifter:
It is important to check for correct "layering" of available sprites / assets. 
This was mentioned on #140 where many of the _dummy assets_ had unpolished collision and all. 

Of course you will keep track of this and implement it well, just to mention :)

mango-gremlin:
closed as everything here is done.

PR #140, opened 2023-12-11
Feature/refactor overworld

### LINKED ISSUE #98: overworld : simplify process of creating new item location on overworld
## Motivation

makes it easier to place and position new items within the world 

## Implementation

We can solve this similar to the bridge-interaction. 

It is containing an interaction area yet has its own exposed values which can then be set in the editor.
Upon **loading the game** it will then forward those information to the interaction area which will update its contained information that are shared upon interacting with it. 

Contained values might be: 
- itemtype, 
- name, 
- description, 
- maybe what is required to get them  ( like reputation of X to interact with them ) 

**things like reputation and more** should be discussed with the **game design team soon**

### LINKED ISSUE #135: overworld | denote item description based on its type
## Motivation

items don't need individual descriptions, they should be **set once** and then used for every time of the given type!

## Implementation

Upon item creation the corresponding description is set, matched on the prepared item type.

ScatteredDrifter:
## Motivation: 

This PR tries to lay a groundwork to continue working with the overworld and to **establish a codebase** we can continue to work. 
It aims to resolve the following:
- resolves #98 
- continues with #133 ( yet here the game design team announced that those limitations are not required or deployed anymore. hence it is more or less resolved **(?)**
- resolves #135 
- began with #136 

I'm aware that its not perfect at all and might have many approaches that could be troublesome, too weird or simply not good design, **please notify me about those** so that I can improve them 

---

## Overview: 
This PR changes many things at once: 
- corrected positions of files within `/overworld/` -> Ui is now located at the supposed spot, entities like the player were moved to the corresponding position
- overhaul of interaction between player / and interaction areas 
- fundamental npc interaction ( yet no dialogue boxes / or complete dialogue system ) 
- improved item interaction ( not just a interactionarea with custom value anymore ) 
- improved bridge interaction 
- several changes in structure of items.

### | Overworld Interaction |
I changed the way the player is interacting with areas within the overworld:

Every point of interest (POI) is denoted as an area the player can enter to interact with it. Those POI include:
- NPCs  ( for talking or quests ) 
- Items ( or areas where one can find an item )
- Bridge level "entries" 

They are denoted and defined as enum in "interactionspot.gd".

This node acts as **foundation** for any specialized interaction and is **instantiated and updated** by any other _more specific_ node. 

for a little overview: 

"interactionspot":
-> instantiated by "item_interaction"  ( sets item, availability, methods to extract and query the item ) 
-> instantiated by "npc_interaction" ( sets npc specific information ( dialogue, quest, rewards and more ) 
-> instantiated by "bridge_interaction" ( sets bridge values ( id, description, possible image to give preview) 

This means that "interactionspot" will always be a child of the more specific parent which contains **the specific information**.
The "interactionspot" will only serve as **query** to gather the information and pipe them to the player at the end. 

Whenever the "player.gd" enters such area it is added to its stack of available interactions. 
For now those can then be "activated" by pressing "E". 

Upon pressing "E" the player getting the newest instance of an interaction. 

**any interaction** contains an enum to denotes its type ( one of the POI above). The _Player_ is then checking for that type and based on it they are doing different things.  ( to be found in "Player.gd" ) 

upon interaction the method "interact_with_area()" contained in "interactionspot.gd" is executed which, based on the selected type", executes several methods from the **more specific node** that either denotes an interaction with "npcs", "items", "bridges". 

**This method will always return** a dictionary containing the queried and received values which can then be processed by the player itself ( either add item, leave scene ... ) 

**within "interactionspot"**
the method "interact_with_area()" is different based on the given "interaction_type" ( one of the POIs): 
for **items**:
- the item is queried 
- the dialogue is queried ( can either be success - item obtained - or failure - item not obtained due to reason X - ) 

for **Bridge**:
- check whether it was solved 
- gather bridge_game_id 
- gather description of level ( or information that it was solved already ) 

for **NPC**: 
> _this one is not done yet

- gather dialogue ( depends on state of interaction ) 
- receive value ( in case quest was successful ) 

**Question:** should we add the upper explanation into the wiki to give a background on how the interaction system works?

---

## Items:

An Item is an object of type "Item". they have a "custom description" yet we only have a limited amount of items in our game " which should have a single description per type. 

to simplify the process of having a description for each type, I rewrote "obtain_item_description" to simply map the description, **whenever it might be used** for visualization or similar. 

--> 

The inventory currently stores any object of an item individually: 
Those instances might be "trivial" because they are not necessary / dont have to be an object. ( this is also referenced in #139 ) 

**Because** I changed the way items are represented / used, I also changed the "item_interaction" to comply to the new changes. 

## Testing
The overworld scene now contains example interactions where one can test the new features: 

**Dialogues and solving a quest** are not operational as of now because I have not found a elegant way to give an item to an npc **(and it will hopefully be replaced / improved when focusing on the dialogue system)**


ScatteredDrifter:
I will get to the conflicts in a minute

koennixer:
The new interaction is functional and easy to apply. Calling interaction with "E" works. Nothing seems to break (at least more than previously). On first glance solid. I will leave a deeper comparison to Simon who is more familiar with the code.

### Potenial Problem 

I don't know if this is intentional but the bound does not work:

outside bounds

Also the text is wildly out of bounds on some interactions:

outside bound text

Later on we will change it to a text box anyway, but even then formatting the next to size is important, so implementing the text being formatting according to a bound now for the "unfinished" version just means we have a solid ground work for the "finished" version.

### Layers
This is not caused by this change, but while we are doing a general overhaul we might wanna address some weird layering issues:

layers!
layers2!

ScatteredDrifter:
Thanks for your review ^^ 

Many of the mentioned issues are primarily due to missing components ( i.e. sample level, missing dialogue field and system ) ( I would argue ) 

> I don't know if this is intentional but the bound does not work: 
Well yes its somewhat intentional because no real effort was put into actually creating / providing a self contained ( closed) island to showcase. 
We do have an issue tracking this however #57, and its on plan for the upcoming sprints ( until january!) 

> Later on we will change it to a text box anyway, but even then formatting the next to size is important, so implementing the text being formatting according to a bound now for the "unfinished" version just means we have a solid ground work for the "finished" version.

While I do agree with your intent / aspects I would amend that: 
- this is purely to show that the dialogues are shared correctly between player and interaction 
- pre-formatting the text might not be the best location for this process ( and should instead happen in the text-box that is given this string. ( I'm sure there are options to **auto-wrap** text according to available space. 

> This is not caused by this change, but while we are doing a general overhaul we might wanna address some weird layering issues:

Valid this will be worked on once we implement a tileset with **self-made assets** , also mentioned it in #119  once more

Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -136,10 +147,18 @@ func check_input():
 
 # adding item to first position of inventory
 func add_to_inventory(item:Item):
-	inventory.insert(0,item)
-	# emit signal to update Ui
-	updated_inventory.emit(inventory)
+	if item.item_type == Item.ItemType.NONE:
+		# received a value that is not valid 
+		return
+	else:
```
DerSimi, line 153: xD

DerSimi (approved):
All in all, it's just fine. The interaction spot is well designed.

Arch-enemies/overworld/entities/Player/Player.gd
ScatteredDrifter: Well its useful because it allows for this "NONE" to denote that no item was obtained ( like if the player interacts with an npc and it was not solved yet, we **dont have to return** something like null! but can just pipe this _none item_ which will not be added to the inventory later ( once it percolates all the way to be added to the inventory!)

DerSimi: I meant your else statement, it doesn't seem necessary to me 

Regarding your approach: I like it, as you already mentioned, None ItemType avoids null.

ScatteredDrifter:
> I meant your else statement, it doesn't seem necessary to me 

Ooh I forgot about that yes, I can remove it in a few minutes

koennixer (approved):
Given the additional context I see no issue with merging this. Especially given that you already improved and interrated on this in other issues.

PR #141, opened 2023-12-13
Bug/issue100 | updating inventory ui upon loading save file 

ScatteredDrifter:
## Motivation
closes #100 
loading the config was done by directly setting the inventory with the generated inventory by the save_management module. 
Thus, because it was just set directly, there was no signal emitted that the inventory changed and thus it maintained empty until an item was added / removed. 

## What was changed/added
I added a function `set_inventory` that takes an Array of Items and set this to be the players new inventory. 
After doing so it emits a signal to update the Ui. 

This method is now called upon loading the save file 

## Testing
runs well, not much to test.

PR #147, opened 2023-12-14
changed inventory system #139

### LINKED ISSUE #139: overworld | change inventory system

DerSimi:
## Motivation
closes #139

## Testing
Works. The only thing is, that the honey can only added once (and again after rejoining overworld). I don't belive that this has something to do with my implementation. Because with the stick it works perfectly.

## Additional Information
Inventory is not shown on joining overworld. This issue has already been resolved. See https://github.com/mango-gremlin/arch-enemies/pull/141

### DIFF
--- a/Arch-enemies/overworld/entities/Player/Player.gd
+++ b/Arch-enemies/overworld/entities/Player/Player.gd
@@ -16,7 +16,15 @@ signal saved_player()
 # Array to hold the players useable items 
 # FIXME specifying the type didn't work, tried Dictionary[Item, int], Dictionary{Item, int},
 # Dictionary<Item, int> :/
-@onready var inventory:Dictionary = {}
+
+# FIXME this is the temporary solution to get the base structure
+# in #134 it should be refactored and be available easier / more beautiful xd 
+@onready var inventory:Dictionary = {
+	Item.ItemType.STONE : Item.new(Item.ItemType.STONE),
+	Item.ItemType.LEAF : Item.new(Item.ItemType.LEAF),
+	Item.ItemType.HONEY : Item.new(Item.ItemType.HONEY),
+	Item.ItemType.STICK : Item.new(Item.ItemType.STICK),
+}
 
 # Array to hold hte players progess in the world 
 # used to check whether the user can traverse a certain region
@@ -121,28 +129,6 @@ func execute_interaction():
 			_: #default
 				pass
 
-# uses item and removes its entry from inventory
-func use_item():
-	# TODO could use pattern matching to use the item accordingly
-	if !inventory.is_empty():
-		# inventory not empty, using item
-		
-		var itemToRemove = null
-		
-		for item in inventory:
-			var amount = inventory[item]
-			
-			if amount != 0:
-				itemToRemove = item
-		
-		inventory[itemToRemove] -= 1
-		
-		var used_item = itemToRemove
-		
-		# TODO as long as this is a string, we can use this as debug
-		print(used_item)
-		# update UI
-		updated_inventory.emit(inventory)
 
 # checks against definde inputs, takes action if action was registered
 # TODO naming could be improved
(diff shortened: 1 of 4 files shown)


Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -198,7 +215,13 @@ func save_state():
 	var json_inventory = []
 	
 	for item in inventory:
-		json_inventory.append(item.to_json())
+		var amount = inventory[item]
+		
+		# store amount
+		var itemDictionary = item.to_json()
```
ScatteredDrifter, line 221: naming convention :)!

Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -209,6 +232,16 @@ func save_state():
 		"zoom": $Camera2D.current_zoom
 	}
 	return state
+
+# FIXME: Would it be better to save the itemtype in the dictionary, not the item itself?
```
ScatteredDrifter, line 236: agreed, yes

Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -209,6 +232,16 @@ func save_state():
...
+# This seems to be the only solution, preventing find_item_instance, there is no way of 
+# overriding the equal or hash function of a class to influence the dictionary in godot.
+func find_item_instance(item:Item):
+	for tmp in inventory:
+		if tmp.item_type == item.item_type:
+			return tmp
+	
+	return null
```
ScatteredDrifter, line 244: null values!

ScatteredDrifter: I'm actually not sure whether we need this function at all. 

Whenever we use an item we are going to:
- check whether we have the correct amount available --> then remove it and signal true to the requesting site

Hence we dont really need **any item instance** in our game, because they are not used as such. You can read some considerations regarding that here #149

Arch-enemies/overworld/scripts/Item.gd
```diff
@@ -37,7 +37,7 @@ var item_type:ItemType
 # --- /
 # -- / class constructor 
 func _init(new_item_type:ItemType):
-	item_type= new_item_type
+	item_type = new_item_type
```
ScatteredDrifter, line 40: reasonable ^

Arch-enemies/overworld/scripts/save_management.gd
```diff
@@ -64,13 +64,14 @@ func generate_inventory(inventory_data):
 	print("Generating inventory from string: \"", inventory_data, "\"")
 	# iterate over each item and generate an item from each 
 	
-	var inventory:Array[Item] = []
+	var inventory:Dictionary = {}
```
ScatteredDrifter, line 67: as mentioned above I would simply pre-write this dictionary as its somewhat static anyway, is it not?

Arch-enemies/overworld/scripts/save_management.gd
```diff
@@ -64,13 +64,14 @@ func generate_inventory(inventory_data):
...
 	
 	for inv in inventory_data:
```
ScatteredDrifter, line 69: whats represented with "inv" ? 
What type is inventory_data ?

ScatteredDrifter (changes requested):
Thank you for your contribution, I appreciate your work ^^ 

I've made some comments in the code which should be resolved or considered. 

However I will also push some changes I've done because I think it might be better to implement our inventory in this way. 

--> **Please tell me whether its fine** or not

## Consideration: 

I've mentioned it in a comment already, however I think that we should change the way the inventory is represented / works: 

Consider we are creating the inventory with every item available loaded already at the beginning.
Now we are only going to increase / decrease its value.

That way we can:

    constantly display the amount we have for any given type
    easily update and search for an item --> because we know that its always there!

ScatteredDrifter:
fixed merge conflict in newest commit cf822 
--> this conflict was caused by #141

Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -8,13 +8,13 @@ signal saved_player()
...
 @export var zoomlevel:Vector2 = Vector2(1,1)
-@onready var anim :  AnimatedSprite2D = $AnimatedSprite2D
+@onready var anim:AnimatedSprite2D = $AnimatedSprite2D
 
 # --- / 
 # -- / player states
 
 # Array to hold the players useable items 
-@onready var inventory:Array[Item] = []
+@onready var inventory:Dictionary = Item.init_items()
```
ScatteredDrifter, line 17: way better, thx

Arch-enemies/overworld/scripts/Item.gd
```diff
@@ -126,4 +147,16 @@ static func string_to_item_type(str):
 			# usually not the case 
 			print("Invalid string representation received")
 			return ItemType.NONE
+			
+
+# initalizes inventory with the corrosponding Item instances
+static func init_items():
```
ScatteredDrifter, line 153: this one's great, nice!

Arch-enemies/overworld/scripts/save_management.gd
```diff
@@ -62,18 +62,20 @@ func load_config():
...
-	
-func generate_inventory(inventory_data):
+
+# takes dictionary and 
+func generate_inventory(inventory_data:Array):
 	print("Generating inventory from string: \"", inventory_data, "\"")
 	# iterate over each item and generate an item from each 
 	
-	var inventory:Array[Item] = []
+	var inventory:Dictionary = Item.init_items()
```
ScatteredDrifter, line 71: thanks!

ScatteredDrifter (approved):
Thanks once again for your changes and additional implementations. 

ran it locally to confirm that they are not breaking anything, works well :)

Thanks

PR #151, opened 2023-12-15
Fixed wrong dropzones lighting up

### LINKED ISSUE #129: too many dropzones are triggered from time to time
## Description
Especially for the snake, too many dropzones are triggered, or dropzones that are not touched at all.

**To Reproduce**

**Expected behavior**

**Screenshots**

https://github.com/mango-gremlin/arch-enemies/assets/57258671/ffadd0df-8e97-4477-a3ba-d11cac587f7a

## Solution Proposal
How could we solve this? Leave blank or assign someone if no ideas

Raffael-Rack:
## Motivation
closes #129

When attaching animals to each other the correct zones should light up.

## What was changed/added

On the squirrel and spider the collision zones of the top and right dropzone where switched, so I switched them to their correct position. Additionally collision zones register overlap even if they only occupy adjacent pixels and their edges touch, so I reduced the size of the collision zones by 1 around the edges. This should not lead to any problems since animals only move in increments of 10

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116198748/1b95f440-119b-4611-adb7-462d77e2986e

### DIFF
--- a/Arch-enemies/bridges/scenes/animals/snake.tscn
+++ b/Arch-enemies/bridges/scenes/animals/snake.tscn
@@ -11,13 +11,13 @@ size = Vector2(48, 10)
 size = Vector2(48, 8)
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_lyvfn"]
-size = Vector2(9, 10)
+size = Vector2(8, 8)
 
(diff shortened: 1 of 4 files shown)


Kabooni (approved):
Works well

PR #155, opened 2023-12-15
animals can now only be connected to the shore or to other animals al…

### LINKED ISSUE #128: prohibit dragging of connecting animals
## Description
Animals who serve as a *connecting* piece should not be able to be dragged. This concerns mostly animals which have another animal in their top_dropzone, as well as special rules for the spider

## Implementation
Implement a function that checks all animals (get them by type, or by new group), and checks the position of adjacent animals and add or remove the ``draggable`` group from them. Should either be called inside the process function, or after and before picking up an object. The latter is probably better, so it doesn't cause problems with currently dragged object.

Use dropzones (as they are Area2D), and check with their overlapping areas.

### LINKED ISSUE #132: animal placement should only be possible if the bridge is connected to a shore
## Description
Like the animal connector issue, to have a functional bridge, the animals should be connected to the shore before the player can cross it. 

**To Reproduce**

**Expected behavior**

**Screenshots**

## Solution Proposal
You could simply deny placing animals that are not connected to the shore via the bridge, or not allow them to go into fox mode without a functional bridge.

trashsolid123:
## Motivation
closes #132
and 
closes #128

animals can now only be connected to the shore or to other animals allready on the shore, animals in the middle of the Bridge cant be taken out

## What was changed/added
added 2 variables in the extends StaticBody2D to save the animal/shore an animal is holding on to and if another animal is holding on to this one. When moving an animal it checks if it can move and if it is being connected to the Bridge.
var connected_to = null
var someone_connects_to_this = false

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/44133213/68c8ff38-22ee-4cab-9f8e-21a2c7b0cca9

## Additional Information
var someone_connects_to_this = false

can be changed to a List to store all animals holding on to this one.


Kabooni (changes requested):
The idea is great, but before I can even begin reviewing:
1. Please update to Godot 4.2
2. Merge main into this one, you still have the version before the refactoring

Kabooni (approved):

Kabooni (changes requested):
You changed the beginning scene to our bridge scene (happening in project.godot), that shouldn't be happening

ScatteredDrifter:
considering that  I'm not really involved in this development (( and am not fluent in your implementation either )) I will un-assign me from reviewing this change, sorry )

Kabooni (approved):
lgtm

PR #164, opened 2023-12-20
Forbidden zone of dragged snake is now working

### LINKED ISSUE #127: prohibit placing a snakes head next to other animals
## Description
It should not be possible to place a snake where its head touches another animal, since that is a forbidden zone.
(E.g.: placing a snake ontop of a snake)

**To Reproduce**
1. Go to the bridge building.
2. Place a snake and than place a second snake (shifted one block to the left) on top of the first snake.
3. The game will you let you place it there even though it shouldn't.

**Expected behavior**
The game shouldn't let you place the snake there.

**Screenshots**
grafik

## Solution Proposal
Check forbidden zones of dragged animal as well

aylinfi:
## Motivation
closes #127 

There was a problem where the forbidden zone of the snake that is being dragged didn't prohibit placment.

## What was changed/added
The forbidden zone is now part of a Area2D that makes sure you cannot place the snake while something is inside of the zone.

You can now also see the forbidden zone of the snake that's being dragged, to better understand the rules of placing it.

## Testing
https://github.com/mango-gremlin/arch-enemies/assets/116217918/f1be90d9-6970-49ca-a6fb-80d4a7a3485c

You cannot place the snake in wrong areas anymore.

## Additional Information
grafik

There are two `CollisionPolygon2D` and `ColorRect` that should be completely identical but need to be both the direct child of the `Area2D` node and the `StaticBody2D` node.

I would think there is a nicer way for this to work with only one pair, but I couldn't figure it out.


marcel-iv (approved):
Code looks good, and it doesn't seem to cause any issues that didn't exist before

However, the duplicated nodes do not look great, but thats an issue we had before (see snakeBod/snake), I don't there is any good way to prevent that, at least not that I'm aware of. If anyone finds out a better way to do, we can maybe do a refactor if we have enough time, but it doesn't seem to cause any issues so far.

Kabooni (approved):
It doesn't look nice, but it works, and I don't have an idea how to implement it any better

PR #166, opened 2023-12-21
Implemented functionality which prevents placing animals inside eachother

### LINKED ISSUE #123: Bridges bug collection tasklist
## Motivation
It's important to have an overview of all the bugs currently in the game and future ones we find. This tasklist might never be finished completely - but we can try to do as much as possible.

### Deadline
09.01.

## Bugs we can reproduce
- [x] #126
- [x] #129
- [x] #130
- [x] #131
- [x] #162 

## Bugs we can't reproduce consistently
- [x] #157
- [x] #158
- [ ] sometimes, the grid doesn't appear, and everything breakes
- [ ] one time, when going into fox mode, all dropzones and the grid were always visible

### LINKED ISSUE #130: the main hitbox of an animal should be a forbidden zone
## Description
The inside hitbox of an animal should be a forbidden zone to prevent bugs and weird clipping behavior.

**To Reproduce**

**Expected behavior**

**Screenshots**

## Solution Proposal
How could we solve this? Leave blank or assign someone if no ideas

Raining-Cloud:
## Motivation
closes #130

## What was changed/added

Added a new group `animal_body` to the Area2D which already represents the body of the animal. By this we can check in the code if we have overlapping areas which also are a animal. This is done with a simple for loop in the drag_handler-

## Testing
Tested in multiple combinations which may cause aproblem

https://github.com/mango-gremlin/arch-enemies/assets/59502349/d101fe49-31ac-4c90-9508-e72cfe6f21a7


marcel-iv (changes requested):
Works and doesn't seem to be causing bugs

However, the small segment of code (in the ``is_correct_placement`` function) is for some reason not inside the for loop, so we have two identical for loops, sequentially. The new if clause should simply be put inside the loop, if that doesn't cause any bugs.

 Also, please add a comment to your code, even though its fairly simple to understand.

Raining-Cloud:
> Works and doesn't seem to be causing bugs
> 
> However, the small segment of code (in the `is_correct_placement` function) is for some reason not inside the for loop, so we have two identical for loops, sequentially. The new if clause should simply be put inside the loop, if that doesn't cause any bugs.
> 
> Also, please add a comment to your code, even though its fairly simple to understand.

What kind of bugs? Due to I only added another return false option the Bug must be that the Animal cannot be placed in a certain spot. Otherwise its the general code causing problems. I don't get what you mean by the identically sequential loops?

marcel-iv:
image
I meant these two loops, however, I just realized that your loop checks for ``overlapping_areas``, while the already existing loop iterates over ``overlapping_bodies``. If this is necessary for your implementation, please feel free to accept and merge your PR.

Otherwise, everything looks great, thanks for adding the comment.

As previously mentioned, all bugs encountered while testing your code where there before, as covered in #123. Your approach seems robust, and easy to check for any bugs that we might have missed, which is great.

marcel-iv (approved):

PR #178, opened 2023-12-24
Add bridge tilemap and collisions

### LINKED ISSUE #177: Bridge tilemap
## Motivation
More pretty assets instead of blocks.
This is part of #118

## Implementation
Draw a tilemap. Easy!

Kabooni:
## Motivation
closes #177
Now we can create pretty bridge levels!

## What was changed/added
I've drawn a tilemap and added corresponding collisions.

## Testing
image
These are all existing new tiles, note that we don't need everything all the time of course.

https://github.com/mango-gremlin/arch-enemies/assets/57258671/6a72c033-0a72-4d7e-9dbd-51cb8ebb75b7
Collisions works as you can see. I've left the green background tiles, because we need them to see the grid more clearly. They will be replaced by the background provided by @Raffael-Rack later.


mango-gremlin (changes requested):
looks really good!

I have two suggestions:
Screenshot from 2023-12-24 12-00-20
here there is like this gap in the green where the water lily is - i suppose this is because the water lily tile cannot be in the same position as a green tile. You can get around this issue by making a new layer for the platforms, and for the background. This would be good to have since there shouldn't be gaps in the end. You can make a new layer in the inspector when klicking on the tilemap, under 'layers'

Also, I think for future levels and modularity it would be important to save the tileset in an external file.
This can be done here:
image

Lastly, I see that you put the files for the tilemaps in to the folder bridges/assets
now the question is, i thought we said we would put all assets in the art -> assets folder, so we have all assets in the same place (since some are going to be used for both bridges & overworld. So if we do this, you should move all files to the folder where the other tilemaps are (and change the names so they indicate that these tiles belong to bridges). I thought we agreed on this, but if you want to do it differently, tell me, because i need to then also change my implementation for the overworld.

Kabooni:
I've changed all the requested things, as these are indeed valid points. Note, that saving the tileset in a separate file makes it only slightly more modulated, as the images used in the tilemap can't be saved in the tileset and need to be separately saved in the project as well.

https://github.com/mango-gremlin/arch-enemies/assets/57258671/e17e6108-8bef-4407-94c3-fd06b3b4440c

I've also added some new tiles, in detail
-two tiny frögs
-two mr crabs
-some sea grass with different heights

mango-gremlin (approved):
looks really good, thank you!!

Ah, i didn't know that about saving the tilemaps, thank you for that. That is a bit unfortunate, but i hope we can make it work somehow.
But the tileset file does save the collisions and such so we dont need to redo those at least? i hope?

The frögs and mr crabs are amazing, i love them. Also the seagrass ties everything together!

Kabooni:
> Ah, i didn't know that about saving the tilemaps, thank you for that. That is a bit unfortunate, but i hope we can make it work somehow. But the tileset file does save the collisions and such so we dont need to redo those at least? i hope?
 I think so, or the tileset file would be pretty useless

PR #181, opened 2023-12-25
Bridge Overhaul using a hard grid

koennixer:
## Motivation 
Until now we have used a soft-grid approach were the grid is implied but not actually used. This has caused issues in implementation. This is a trial at a new approach using a hard grid. Details about the functionality are found in the wiki at https://github.com/mango-gremlin/arch-enemies/wiki/03-Bridge-Using-Hard%E2%80%90Grid.

## Testing
Here is a short demo. But trying it yourself might be a good idea.

https://github.com/mango-gremlin/arch-enemies/assets/65815628/c985a695-82ca-49ac-852f-6607da5928b5

## Additional Information
This is very much just the simple groundwork. Much improvement is possible. Many of the features desired should be much easier to implement in this frame work.


aylinfi:
just some observations/questions:
- I couldn't figure out how to place something until I realised you place it where the mouse cursor is pointing to and not the animal itself, maybe the animal could follow the cursor more closely?
- you can't drag animals again after placing them? (I know about the undo button but it makes bridge building more tedious than before)
- you should be able to place the spider in more places (e.g. underneath the snake)
- dragging nothing also activates the hitbox visibility

But overall very impressive, well done!

koennixer:
Just in case some others run into the same observations.

> * I couldn't figure out how to place something until I realised you place it where the mouse cursor is pointing to and not the animal itself, maybe the animal could follow the cursor more closely?

Yes, this is something that can be addressed. It is a bit annoying, because you have to orient the created TRect, but it should be doable. The dragging options already in godot are just inherently janky.

> * you can't drag animals again after placing them? (I know about the undo button but it makes bridge building more tedious than before)

This would be extremly hard to implement, because then each placed animal would have to create a node that can be dragged and each dragging of an animal would have to check if it makes other animals position illegal. I would not recommend implementing that. Personally, I do not think that using reset and go back is much more of a hazzle and has a charm of its own, but that is something for game-design to discuss and deciede.

> * you should be able to place the spider in more places (e.g. underneath the snake)

I was not fully caught up with the logic of how and when objects can be placed. But this is an easy fix, if there are zones that are not FORBIDDEN or ALLOWED but instead CONDITIONAL, then you would just have to add that as an Entity-Type and update some functions to account for that.

> * dragging nothing also activates the hitbox visibility

The dragging currently has the bug that if you place nothing the hitboxes are shown permanently. That is the bigger fix and trying to drag something other can be done as a side fix.

koennixer:
Also added Conditional Zones as mentioned in the comment by [aylinfi](https://github.com/aylinfi).

mango-gremlin (approved):
looks good, we decided we should switch to this implementation.

Issue #158, opened 2023-12-16
Sometimes, a different animal is moving than the selected one


Kabooni:
## Description
Sometimes (we can't reproduce it for now), when clicking and trying to drag an animal, the last one used is moving. Usually, it results in an animal snapping back to (0,0)

**To Reproduce**

**Expected behavior**
It should not happen at all, only the selected animal should be moving.

**Screenshots**

https://github.com/mango-gremlin/arch-enemies/assets/57258671/74622034-125a-4254-a57f-90cfc9c40c01

**Possible clues**
1. After commenting out the content of mouse_exited, it happens every time
2. Retracing the same steps that led to the bug does not repeat the bug every time. So it shouldn't be a detecting/triggering problem
3. I suspect it's not a problem with the code, but something related to file loading, file accessing, or something else of that sort. We might never find the true cause, maybe we should focus on preventing it as far as possible

mango-gremlin:
due to implementing #181 this will no longer be necessary

PR #182, opened 2023-12-26
Fixed snapping bugs

### LINKED ISSUE #157: sometimes, animals do not snap (especially when playing it on top of the snake)
## Description
Sometimes the animals do not snap to positions where they should be placeable when you drag them there, instead returning to their original position after letting go. 

While this happens, often there is the added visual bug of dropzones lighting up even though nothing has been placed in them.

**To Reproduce**
(in the video below, i did the following)

1. drag snake somewhere
2. place squirrel on snake, next to forbidden zone (not in it)
3. try to place spider in the zone to the right of the squirrel (top) -> the zone below lights up, which is in the forbidden zone around the snakes' head, and the spider doesnt want to be placed, snapping back to its original position.

**Expected behavior**
Stuff should snap where you drag it, if it's allowed. In the above example, the spider is allowed to be in the zone.

**Screenshots**

https://github.com/mango-gremlin/arch-enemies/assets/104830903/79c31c63-ea21-4253-ad3f-8a5e93912b4e

This shows wrong dropzones lighting up with every animal i want to place, which is propably part of the reason for the bug.
As you can see around second 20, the dropzone below where i want to place the spider lights up, and the spider does not get placed where I want to place it.

## Solution Proposal
Solve the problems of wrong dropzones lighting up 
- make sure only the dropzone the currently dragged animal touches lights up 
- this should solve the problem because now forbidden zones will only be triggered when there is actually something in them.

marcel-iv:
## Motivation
closes #157

## What was changed/added
The hitbox of the snake head was adjusted to prevent bugs concerning falsely detecting collisions with it. Likewise, all central animal hitboxes have been adjusted, preventing bugs concercing falsely detecting collisons with itself. Additionally, ``is_inside_dropable`` was removed, as it served so purpose in the current approach. The functionality from #164 (which was deleted by #166) was re-added and greatly improved, and shouldn't cause bugs now. I also removed a duplicated (and non-functional) check for forbidden zones, as this only causes confusion. Some print statements were added to aid future debugging.

## Testing
Every placement works as intended, however the bug mentioned in #158 is predicatably still there.

[everythings works no bugs](https://github.com/mango-gremlin/arch-enemies/assets/104799849/9a4a3401-0fa0-4234-838e-96efb6848a05)

Here you can see that the snake's head works again.

[forbidden zones work](https://github.com/mango-gremlin/arch-enemies/assets/104799849/6a77e1bc-7d30-40dd-a49a-b60f6231d120)

Feel free to test for more snapping bugs in my branch, just in case I've missed something. 

## Additional Information
If PR #181 is implemented, this PR should likely be closed, as it fixes bugs in the current approach only.

Originally, I planned a larger refactor of the ``is_correct_placement`` function, but as this might become obsolete if #181 is implemented, I decided not to work on this for now. Regardless, the code should be easy enough to understand as is.

### DIFF
--- a/Arch-enemies/bridges/scenes/animals/snake.tscn
+++ b/Arch-enemies/bridges/scenes/animals/snake.tscn
@@ -50,7 +50,7 @@ script = ExtResource("3_fqrgl")
 
 [node name="CollisionPolygon2D" type="CollisionPolygon2D" parent="forbidden_Area2D/forbidden_dropzone"]
 position = Vector2(-50, 11)
-polygon = PackedVector2Array(40, -26, 40, -17, 50, -17, 50, -5, 40, -5, 40, 4, 59, 4, 59, -26)
+polygon = PackedVector2Array(40, -25, 40, -17, 50, -17, 50, -5, 40, -5, 40, 3, 58, 3, 58, -25)
 
 [node name="ColorRect" type="ColorRect" parent="forbidden_Area2D/forbidden_dropzone"]
 z_index = -1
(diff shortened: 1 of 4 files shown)


mango-gremlin:
due to implementing #181 this will no longer be necessary

PR #197, opened 2023-12-27
Added pause-menu to bridge

### LINKED ISSUE #174: bridges | embed pause menu from overworld into bridge level
## Motivation
You should be able to access the pause menu from everywhere in the game.

## Implementation
**It is important that the main menu which was already implemented in the overworld is the same menu that is accessed through the bridges. You should not simply copy-paste everything, but refer to the same scene.**

You can find the already implemented pause menu here:
```
Arch-enemies/overworld/ui/menu/menu
```

The menu consists of the script `pause_menu.gd` and the scene with the same name. 

- when you are in the bridge level, you should be able to press `ESC` and the pause menu appears 
-> to see how this was done in the overworld, check out how they did it and just do the same thing

**UPDATE**
In the last high-court meeting it was decided to embed the pause menu directly into the current scene, and not open a new one. It should contain all the old options, and make it a lot easier to implement a return button to the current active scene.
This could be done similarly to the goal menu, but should be generic enough to be able to be reused in the overworld without changes.

aylinfi:
## Motivation
closes #174 

The pause-menu in the overworld should also be in the bridge building scene.

## What was changed/added
I followed the same structure as in the overworld Player.gd to call the pause-menu scene.
Now you can open the pause-menue by pressing `esc` while in fox-mode.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/7d4c6804-17af-413d-ac98-29a31cd5fecd

## Additional Information
When opening the pause-menu you cannot go back to bridge building. This can either be fixed by changing the pause-menu scene to include an extra button (but what's the point of this button while pausing in the overworld?) or by creating a seperate pause-menu scene for the bridge building.

There is also the question if we should save the bridge build when using the pause-menu or not.

### DIFF
--- a/Arch-enemies/bridges/script/player.gd
+++ b/Arch-enemies/bridges/script/player.gd
@@ -44,6 +44,10 @@ func _physics_process(delta):
 	else:
 		velocity.x = move_toward(velocity.x, 0, speed)
 	
+	# pressing esc -> pause-menu
+	if Input.is_action_just_pressed("open_menu"):
+		enter_pause_menu()
+	
 	move_and_slide()
 	update_animation()
 	update_facing_direction()
@@ -99,3 +103,15 @@ func _on_animated_sprite_2d_animation_finished():
 		animation_locked = false
 	elif animated_sprite.animation == "jump_up":
 		animated_sprite.play("jump_fly")
+
+
+func enter_pause_menu():
+	print("pause menu")
+	save_bridge()
+	# TODO ? (overworld team not happy with this but why?)
+	get_tree().change_scene_to_file("res://overworld/ui/menu/menu/pause_menu.tscn")
+
+# saves placed bridge 
+func save_bridge():
+	# TODO save what animals have been placed where 
+	print("save placed bridge")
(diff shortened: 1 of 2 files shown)


Arch-enemies/bridges/script/player.gd
```diff
@@ -99,3 +103,15 @@ func _on_animated_sprite_2d_animation_finished():
 		animation_locked = false
 	elif animated_sprite.animation == "jump_up":
 		animated_sprite.play("jump_fly")
+
+
+func enter_pause_menu():
+	print("pause menu")
+	save_bridge()
+	# TODO ? (overworld team not happy with this but why?)
```
ScatteredDrifter, line 111: Primarily because we may want further routines to run upon entering pause menus ^^ 

Furthermore We don't like that the pause menu is an extra scene, and would likely change this in near future

ScatteredDrifter (commented):
Thx for your contribution ^^ 

Took a look at your implementation which seems to be fine so far. 
Considering your question / thoughts: 
> There is also the question if we should save the bridge build when using the pause-menu or not.

We are considering changing the appearance of the pause menu ( to be within the scene instead of a new one being displayed ) which would not disturb the active scene currently used ( like not unloading all the nodes and so its state ) hence with this approach it would not be necessary to save the state.

aylinfi:
Closed by updated version #271

PR #201, opened 2023-12-28
Added autoload for the grid

### LINKED ISSUE #200: bridges | make the grid dependent on the tiling
## Motivation
Currently the grid generates the coloring of the tiles. The other way around is better for level design. In essence the grid should fill the internal representation based on what is inside the tiles.

## Implementation
Change the color_grid and ready based on the expectation that the grid is filled and fill the tileset accordingly.

### LINKED ISSUE #203: bridges | fixed cursor position
## Motivation
Currently the cursor is placed somewhat awkwardly, when it should always be in the bottom left corner. This is only a problem on the y-axis on the x-axis it is automatically correct.

## Implementation
Simply add a control node above the preview in grid_drag.gd and then change the relevant position.

koennixer:
## Motivation
closes #200 
closes #203

## What was changed/added
The ready function not check the tiles and assigns the content of the tiles based on what the tile is filled with. The color function only sets the background to transparent again.

## Testing
This is not extensively tested. A test level should be made to see if it fits specifications.


koennixer:
Now this also closes #203 

It's four lines of codes not worth it's own branch. The solution is just as described in the issue. A control node in set_drag_preview.

Kabooni (approved):
Looks good, I've started my issues on a copy of your branch and I didn't have any major problems that I couldn't solve. But because of that I would say we merge this branch into main, and any required bugfixes can be done on my branch.

PR #209, opened 2023-12-29
Implement mode toggle buttoon and refactor the level

### LINKED ISSUE #202: bridges | implement toggle button (fox/build mode)
## Motivation
Similar to the previous implementation, we need a button to switch between the build and fox ("walk") mode.

## Implementation
Should be the same as our last approach.
I will probably use one branch for this and #204, as I will need some functional platforms for testing purposes.

### LINKED ISSUE #204: bridges | Reorganize level
## Motivation
To have a playable level, we need to visually reorganize the level and
- draw the finished level
- reposition all UI elements already there
- zoom the camera more in, as the grid is too small right now
- make the window size dynamic

Kabooni:
## Motivation
closes #202 and closes #204

Basically, two things are being addressed here:
- adding functionality to the mode toggle button between fox and drag mode
- add all the little things needed to make the level more beautiful and playable

## What was changed/added
- Added code to the mode toggle button, mainly
  - you can walk in fox mode
  - you can't place something or reset placements in fox mode
  - grid and allowed/forbidden zones are only visible in drag mode, while dragging
- I draw a new dummy level that will be changed after the final level design is provided
- Because of that, I repositioned all nodes into the new level and created a camera to zoom in appropriately. This is braking the offset between animal and mouse arrow, but was fixed by @koennixer. I explain the problem later here
- I've also changed the window scaling mode. Now the viewport is always as big as the window, with the default window size like before 
- Continuing that, I've added a water collision, world edge collisions, tile collisions and animal collisions
  -> You still need to hardcode the shore dropzone and fox starting area, but this is not really part of these issues here and could be addressed later while extending the shore dropzones by @marcel-iv. My idea is to be able to draw them on a separate layer and fill their grid cell similarly like the ground and water tiles. 

I hope I didn't forget anything, as I did a lot of little changes.

For the mouse offset problem:
Ideally, we could just use the camera to zoom into the level, and everything would work. In real life, the build-in drag-and-drop functions are tracking the movement without considering any zooming, and this results in movement zooming in unexpected ways, like seen in the video below.

https://github.com/mango-gremlin/arch-enemies/assets/57258671/4828926e-a236-4cda-85f7-f296dde081b4

This is a known godot bug for years and will probably not be changed. But we can also not change the build-in functions ourself, so we either have to code our own drag-and-drop functions, or reposition the animal like @koennixer did.
## Testing
https://github.com/mango-gremlin/arch-enemies/assets/57258671/b782a261-e7a5-4dd3-9b1d-f8a76ccfaba3

## Additional information
I solved all drag mode things like in the old implementation with the global script. We could change that and try an approach with signals, but this would be too much for this pr.
Also the bug while trying to redrag an animal is still happening.


koennixer (approved):
Since I did some tinkering I'm familar with the implementation. It works very well. And can be merged without problem.

I should comment briefly about my fix in regards to the position of the animals when dragging. As it turns out the offset is only impacted by the HBox in which the animal TRects are, so zoom has actually no impact. If we every us a different way to organize the buttons (or add more) we might have to change the offsets again, you could probably also generalize this. In summary this is something to keep an eye one but it can be optimized easily in the future and is currently functional.

marcel-iv (approved):
Fox Mode works exactly as it did before, which is great. The level also looks pretty nice, especially since zoom now finally works, and the preview of dropzones is a pretty cool feature. Code looks well, though I cannot really comment on that, since I have not looked at it in any greater detail. 

However, I noticed that the fox cannot jump over a deer, if its on the same level, and the animal "spawn buttons" for snake and spider aren't aligned to the grid (which isn't important at all, but would look a bit better). This could be fixed before merging

Kabooni:
> However, I noticed that the fox cannot jump over a deer, if its on the same level, and the animal "spawn buttons" for snake and spider aren't aligned to the grid (which isn't important at all, but would look a bit better). This could be fixed before merging

Right now, the fox can only jump three blocks high, but this can be increased if we decide that the player should be able to jump on a deer without help.
I agree with the spawn buttons, but as these are subject to change in the near future (when implementing the inventory), I would let them be as they are. It's just unnecessary work, I think.

Issue #199, opened 2023-12-28
bridges | change drop zones: side / bottom instead of conditional


koennixer:
## Motivation
Currently the drop zones are ALLOWED, FORBIDDEN and CONDITIONAL. The CONDITIONAL option should be changed to be SIDE and BOTTOM. Since some animals can be attached from the SIDE and/or BOTTOM.

## Implementation
Entity.Type SIDE and BOTTOM has to be added.
The code in gird.gd and grid_drag.gd has to be changed to account for SIDE and BOTTOM. This affected mostly update_grid and is_allowed.

mango-gremlin:
additionally, in #211, @marcel-iv mentioned:

> @aylinfi, for issue https://github.com/mango-gremlin/arch-enemies/issues/199, simply edit the lines deciding where dropzones should be tagged. Currently this is 104 and 106. Every other part of the code has been written with the distinction in mind.

PR #211, opened 2023-12-30
Automatic assignment of shore dropzones

### LINKED ISSUE #210: bridges | add all shore dropzones
## Motivation
Some shore dropzones are missing in the new implementation. There should be a dropzone on each shore, on top, and on the side.

## Implementation
In the new implementation from #181, allowed zones have a specific type. 

However - you cannot simply hardcode the position of the shore dropzones, since the grid is loaded from the tilemap during runtime. So you need to find a way either to infer the position of the shore from the loaded tilemap, or if that doesnt work, we need to decide to always place the shores in the same position. 

in #209 @Kabooni had the idea that you draw the shore dropzones on a different layer of the tilemap which makes it possible to dynamically read them out for each level ->  draw them on a separate layer and fill their grid cell similarly like the ground and water tiles. While doing that, it would be nice if the fox starting area, which would be a forbidden zone, would be implemented dynamically like that as well.

marcel-iv:
## Motivation
closes #210 

## What was changed/added
Shore dropzones are automatically assigned, via the following procedure.
During ``_ready``, whenever iterating over a square belonging to a ground tile, check if its adjacent squares (top, right, bottom, left) are in the air (in other words, not ground or water), and within grid bounds (to prevent errors). If they are, add them to their respective list (``shore_top``, ``shore_side``, ``shore_bottom``), depending on their relative position. After the grid has been initialized, iterate over these list and assign them their respective ``ENTITY_TYPE``. As the grid is filled slowly, and not initialized, it was not possible to assign them in the for loop (cells didn't exist, and code currently allows for overwrites), hence the added function.

## Testing
As you can see, all dropzones are rendered and work as expected, without hardcoding them. Note that I edited the Tilemap for this, to show that overhangs also work. The tilemap edit was only done for testing, as is not included in the commit.

[Automatic Assignment of Shore Dropzones](https://github.com/mango-gremlin/arch-enemies/assets/104799849/f0a078be-8a41-4cd0-8831-d5f81ea5fe90)

## Additional Information
I decided against using a tilemap based approach, as this would likely interfere with the current assignment of the grid, and because this approach reduces work required for future levels, by making dropzones automatic.

Also, one could use this code as a basis for implementing hazard zones, by checking adjecant tiles to hazards and flagging them accordingly.

It should be noted that this includes merely the dropzones, and not the initial zones of the fox. This should be done in a future issue, which could also adress some refactoring regarding the initial position (setting position as property, and computing zones from there, for example). 

## Future improvements
I propose splitting the tilemap in two distinct tilemaps; one for ground and one for water. This would require some minor refactoring, but importantly get rid of the hideous hardcoding in line 21 onwards, which is bound to cause some trouble if someone ever edits the tileset in the future. 

@aylinfi, for issue #199, simply edit the lines deciding where dropzones should be tagged. Currently this is 104 and 106. Every other part of the code has been written with the distinction in mind.

### DIFF
--- a/Arch-enemies/bridges/script/grid.gd
+++ b/Arch-enemies/bridges/script/grid.gd
@@ -37,11 +37,9 @@ var fox_start = [Vector2i(1,9), Vector2i(2,9), Vector2i(3,9), Vector2i(4,9),
 Vector2i(1,8), Vector2i(2,8), Vector2i(3,8), Vector2i(4,8),
 Vector2i(1,7), Vector2i(2,7), Vector2i(3,7), Vector2i(4,7)]
 
-# top shore dropzones (allowed)
-var shore_top = [Vector2i(0,12), Vector2i(1,12), Vector2i(2,12), Vector2i(3,12), Vector2i(4,12), Vector2i(5,12),
-Vector2i(33,12), Vector2i(34,12), Vector2i(35,12), Vector2i(36,12), Vector2i(37,12), Vector2i(38,12)]
-# side shore dropzones
-var shore_side = [Vector2i(6,13), Vector2i(6,14), Vector2i(32,13), Vector2i(32,14)]
+var shore_top = []
+var shore_side = []
+# = [Vector2i(6,13), Vector2i(6,14), Vector2i(32,13), Vector2i(32,14)]
 
 #This is the signal we use to transfer the current grid to child nodes
 signal current_grid(current_grid)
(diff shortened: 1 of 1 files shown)


koennixer (approved):
This functions very well. The only improvement feasible is only checking the squares for drop_zone that are likely to have one, e.g. the tile with grass is unlikely to have a drop zone beneath it. But this is a lot of effort and highly unnecessary. As it stands this is perfectly useful and easy to read. If optimization is necessary it can be done at a later stage.

marcel-iv:
> The only improvement feasible is only checking the squares for drop_zone that are likely to have one, e.g. the tile with grass is unlikely to have a drop zone beneath it.

I briefly considered this approach, but decided against it. It would certainly make the code easier, and save a bit of computational power (negligable in our small grid), but would involve some hardcoding, and make the code (and the level visual design) more restrictive, and probably cause issues if we decide to update our tileset. Further splitting our tileset could help this problem, but I'm not sure if this is worth the effort.

PR #215, opened 2024-01-02
Issue213/auto fox startzone

### LINKED ISSUE #213: bridges | add forbidden zone around fox automatically
## Motivation
Like #210, adding zones by hand is tedious and an obvious code smell. To prevent this, and make creating future levels easier, this should be done automatically.

## Implementation
Export the starting position of the fox, and turn adjacent squares to ``FORBIDDEN`` during the ``_ready`` function.

marcel-iv:
## Motivation
closes #213

## What was changed/added
4x3 forbidden zone around the fox starting position is now automatically assigned. This is done by getting the grid square that is nearest to the fox central position, and assigning ``FORBIDDEN`` to all adjacent squares in a 4x3 box (with the central square being 1,2). The conversion from global position in pixels to grid position can lead to the forbidden box being a bit off the actual sprite. As there's nothing one really can do against that (at least not with a simple fix), just drag the fox to a spot where it makes sense. 

## Testing
Here we can see the forbidden box without hardcoding anything.
image

Here we can see one example where the box is a little bit off due to conversion. In this case, drag the fox a bit up or down. 
image

## Additional Information
I mentioned in #213 that exporting the fox' starting position might be good idea, but now I think that simply dragging the fog within the scene is easier for creating levels. If you want to change that, feel free to do so, that takes like 2 minutes.

### DIFF
--- a/Arch-enemies/bridges/scenes/Grid.tscn
+++ b/Arch-enemies/bridges/scenes/Grid.tscn
@@ -681,7 +681,7 @@ scale = Vector2(0.5, 0.5)
 text = "One Step Back"
 
 [node name="Player" parent="Grid" instance=ExtResource("12_y7gdk")]
-position = Vector2(29, 85)
+position = Vector2(30, 82)
 
 [node name="water_detection" parent="Grid" instance=ExtResource("14_yxqwc")]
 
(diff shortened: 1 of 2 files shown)


Kabooni (approved):
Simple solution and works well. The only thing I would propose is to refactor the fox area code into it's own function, to increase readability

PR #216, opened 2024-01-02
Separated CONDITIONAL into SIDE and BOTTOM

### LINKED ISSUE #199: bridges | change drop zones: side / bottom instead of conditional

aylinfi:
## Motivation
closes #199

Before adding the squirrel and its ability it is necessary to differentiate between a side-conditional and a bottom-conditional.

## What was changed/added
In `grid.gd` and `grid_drag.gd` the ENTITY_TYPES.CONDITIONAL has been split up into SIDE and BOTTOM and assigned accordingly in each instance.

## Testing
The bridge building works just like before, the real difference will be visible when the squirrel is implemented.

## Additional Information
Right now SIDE and BOTTOM zones have the same color, but we could change that if desired.

### DIFF
--- a/Arch-enemies/bridges/script/grid.gd
+++ b/Arch-enemies/bridges/script/grid.gd
@@ -45,7 +45,7 @@ var shore_bottom = []
 signal current_grid(current_grid)
 
 #These are the different kind of object we can have in grid cells
-enum ENTITY_TYPES {GROUND, WATER, ANIMAL, FORBIDDEN, ALLOWED, CONDITIONAL, AIR}
+enum ENTITY_TYPES {GROUND, WATER, AIR, ANIMAL, FORBIDDEN, ALLOWED, SIDE, BOTTOM}
 
 func _ready():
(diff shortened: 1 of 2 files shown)


koennixer (approved):
This works. We could consider adding a specific coloring for BOTTOM to separate it from SIDE. But this is not urgent.

marcel-iv:
Looks good. We have discussed the issue of coloring, and have decided that every tile that could be classified as conditional should be orange/yellow (which means we keep the current implementation), as this reduces visual cluttering.

PR #220, opened 2024-01-03
Added squirrel again to new grid approach

### LINKED ISSUE #207: bridges | re-implement squirrel
## Motivation
Squirrel is missing in new implementation from #181 

## Implementation
Same as the other animals, see code in #181, as well as wiki. ask @koennixer , if not sure how.
The squirrel implementation needs #199 to be done first (change conditional dropzones to side and bottom) - or both issues can be done in the same pr.

aylinfi:
## Motivation
closes #207 

Squirrel was missing from new grid approach, so it needed to be re-added.
Squirrel can cling to SIDE zones, so #199 had to be implemented first.

## What was changed/added
Added squirrel related code in all places where other animals where implemented.

## Testing
grafik


marcel-iv (approved):
Code looks good and works just as expected, should be merged into main.

Thanks for also adding the squirrel in such a way that placement is always possible where you would expect it, i.e. not only in the bottom-left corner.

PR #223, opened 2024-01-04
Issue208/re implement goal area

### LINKED ISSUE #208: bridges | re-implement goal area and victory screen
## Motivation
Goal area is missing in new implementation.

## Implementation
The goal area should behave in the same way as in the old implementation:
- if the fox reaches the other side of the level, it should activate the victory menu that can send you back to the overworld.

It could be implemented in the same way than before.

trashsolid123:
## Motivation
closes #208

Implemented in the same way as before.

## What was changed/added
i put the Player in the group "player" and check if the Player entered the GoalArea2D.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/44133213/93e44d3f-8d20-4eb2-be67-00d97859e804

### DIFF
--- a/Arch-enemies/bridges/scenes/Grid.tscn
+++ b/Arch-enemies/bridges/scenes/Grid.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=23 format=3 uid="uid://bl1jporj7hosu"]
+[gd_scene load_steps=24 format=3 uid="uid://bl1jporj7hosu"]
 
 [ext_resource type="Texture2D" uid="uid://d0o2q6fijmd1" path="res://assets/art/tilesets/bridge_tileset/bridge_background_color.png" id="1_uowbf"]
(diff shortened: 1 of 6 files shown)


Kabooni (changes requested):
It works as intended, but for future use, I would suggest some changes:

- After clicking the retry button, it would be nice if the level itself and not only the fox position would be reset. You could use reset_grid() of grid.gd(). I know this was not part of the old implementation, but I don't think it would take a lot of time to do this.
- I'm not 100% sure if it would result in a lot of work, but is it possible to refactor all goal area and menu-related things into a single separate scene? it would make building new levels a lot easier for people not knowing the code.

If something is not doable, just let me know and I can approve it as it is.

trashsolid123:
- the reset button now also resets the grid.
-I put the goal in its own scene. 
https://github.com/mango-gremlin/arch-enemies/assets/44133213/496790dc-fca1-4d14-a0ae-1780932082c7

Kabooni (approved):

PR #226, opened 2024-01-05
Issue 33. Added a Bramble Hazard

### LINKED ISSUE #33: bridges | hazards don't allow animal placement & kill fox
## Motivation
The hazards should also prevent you from placing animals there in one way or another.

Before this can be completed, **hazard assets are needed**

## Implementation
Hazards should do stuff in drag and in fox mode.

They should have the following properties:
- **in drag mode:** With the new implementation from #181 the forbidden zones are now tiles of the type FORBIDDEN. No animals can be placed on a tile of that type. Hazard tiles should be categorised as forbidden.
- **in fox mode:** track whether the fox touches a hazard. Do not simply track whether the fox is in a hazard tile. The hazard may have a complex form that is not a square, and as such the player should only be punished when actually touching the hazard, not every time they enter the a tile that contains part of a hazard. So: find a way to track collisions with tilesets in a way that a tile collision can emit a signal to the fox when the fox collides with it. This has already been done with the water tiles in #209 so it can be done the same way with the hazards.

Raffael-Rack:
## Motivation
closes #33 

We can use hazards to create more interesting levels.

## What was changed/added

1. Drew and added a tileset for hazardous bramble.
2. changed the Grid script to add bramble into the game from the tilemap
3. Implemented hitboxes for the bramble that reset the player like the water hazard.

## Testing

Demonstrating the harmful capabilites of bramble:

https://github.com/mango-gremlin/arch-enemies/assets/116198748/465e1c09-19a1-411c-a5ce-f40b93c4b8a0

Placing animals inside is not possible:

https://github.com/mango-gremlin/arch-enemies/assets/116198748/772b164b-020e-43c7-979f-e77d6d0491a8

## Additional Information
Do you have any feedback on the assets? should they be altered?


Kabooni (changes requested):
The assets look awesome, I love it! The code is also great.
One thing I would suggest is to add forbidden zones on the active layer (1) in all cells where the hazards are, so the player has a visual clue where animals can and cannot be placed. I'm thinking of something similar to the shore dropzone code. Should you have no clue how to do this (or not the time), let me know.

marcel-iv (approved):
The hazard looks very nice, and works as expected. 

I've added some code that adds hazard tiles to `FORBIDDEN` instead, which enables the preview, as suggested by @Kabooni 

Another suggestion would be to move the hazard detection into its own scene, adjust its hitbox a bit (8x8/10x10 square), and instantiate it on all hazard tiles during the `_ready` function. This way, adding hazards would be even easier, but I'm leaving it up to you if you want to implement it in this issue, or open another one for it. 

Edit: I think potentially including automatic placement of area2ds is better done in a seperate issue, the water collision area would also benefit from this change.

Kabooni (approved):

PR #228, opened 2024-01-05
Issue 227 Background

### LINKED ISSUE #227: bridges | nice background
## Motivation

A nice background improves the look of the level.

## Implementation

The basic idea is to create a parallax map with only a single shifting background layer, framed by static side parts.

Raffael-Rack:
## Motivation
closes #227 

A background is very nice to look at.

## What was changed/added

I added a parallax background behind the level.

## Testing

Screenshot Issue 227

## Additional Information

I haven't implemented independent endless scrolling with multiple layers yet, although the assets are designed with that in mind.


marcel-iv (approved):
The background looks excellent, and I like that you make it modular

Kabooni:
After updating this branch from main, I had two change two things:
- Reposition the player, because for some reason the fox was now under the branch drawn in the background
- If you landed directly on a hazard, it could not detect it, because there are two (?) collision shapes and the right one didn't trigger. I'm not entirely sure what the problem was, but making the collision shape a bit larger fixed it.

Issue #219, opened 2024-01-03
bridges | implement hierarchy for tiles


koennixer:
## Motivation
Currently we have three tiles that allow animals placement of some sort. This can lead to conflicts when one square is SIDE and and could also be ALLOWED. This should be fixed.

## Implementation
Establish the simple hierarchy of ALLOWED > SIDE > BOTTOM and adjust the conditions. Best write a separate function for this that is called to make the code functional.

PR #229, opened 2024-01-06
Added deer placement rules, and shallow water

### LINKED ISSUE #218: bridges | implement deer placement rules
## Motivation
We currently have an implementation of the deer, but we haven't implemented its special rule yet: you can place it in shallow water. 

## Implementation
In the initial distribution of Entity Types, check all water tiles: if it has air above it, and ground beneath, mark it as SHALLOW (or SHALLOW_WATER). The deer should be able to be placed inside it. The Enums in `grid.gd` and `grid_drag.gd` should be updated.

## Additional Information
This should  be done after #217, as this depends on the initialization of the grid, to prevent merge errors. 
Also, check if SHALLOW might be overwritten when placing something directly on top of it. I'm not unsure if that can actually happen (I don't think it is), but if that happens, it would permanently disable all shallows once filled with another tile type.

marcel-iv:
## Motivation
closes #218 

## What was changed/added
Added a new entity type - `SHALLOW` - which a water tile with ground beneath, and nothing (or decoration) above it. This is added together with other dropzones in the `_ready` function, and like other types of conditional zones, is colored yellow. 

Implemented the deer placement rule (which we discussed a while ago), which is as follows: The deer can be placed on top of any other animal as normal, and inside shallow waters, as long as the entire deer is within water. 

## Testing
As you can see, you can place the deer only if its either partially on ground, or entirely in shallow waters.

https://github.com/mango-gremlin/arch-enemies/assets/104799849/c5202c3b-1b7c-43f1-bb51-0fed7071c691

## Additional Information
Let me know if you find any bugs or want to change something. If you have any suggestings on the placement rule (not its implementation), that should probably discussed in the discord server. 

Also, in case any merge conflicts arise as a consequence of other merges, I will try to fix those before merging into main.


Kabooni (changes requested):
I've found a bug, right now you can't place a deer in shallow water that overlaps the shore dropzone.

https://github.com/mango-gremlin/arch-enemies/assets/57258671/e3f1e7e9-bd20-418e-827e-fa9dd9bdef8d

**Update**
This seems to be connected to a larger problem and might need a separate issue, as this will require parallel work on #219 and drag-and-drop. As the code works besides that, I will approve it.

Kabooni (approved):

marcel-iv:
the problem is fixed for the deer, but persists with other animals, should be fixed in another issue as discussed

https://github.com/mango-gremlin/arch-enemies/assets/104799849/21a0c29b-9471-4d9d-b9ea-fe959a8b9d3d

Issue #232, opened 2024-01-06
overworld | add assets for npcs in overworld 


ScatteredDrifter:
## Motivation

NPCs are denoted by some animal which we need to visualize in our project! 
Currently we dont have separate sprites for npcs that are _within the overworld_ yet for #224  we do require them! 

## Implementation

add the new assets and replace the dummy values from the new implementation ( will be linked once done! )

ScatteredDrifter:
Hah! I'm stupid and did not see `assets/art/characters/...` where every character is contained without background already, sorry! 

Will therefore close this issue, my bad

PR #233, opened 2024-01-06
overworld | autoloading sprites for npcs

### LINKED ISSUE #224: overworld | autoload npc interaction sprite based on npc_type
## Motivation

Its easier to automatically set the sprite to display for a given Npc-Interaction **based** on the npc-type ( so the animal it is ). 
Furthermore adding an invariant should be easy as we may want a specific npc to have another sprite.

## Implementation
This can be done upon initialization of the npc_interaction itself  and should simply update sprite of this interaction. 
**Furthermore** it should be possible to set a custom sprite which is then **chosen / preferred**  instead.

ScatteredDrifter:
## Motivation
resolves #224 
~~**requires** #232 !~~ was incorporated already, sorry!

## What was changed/added
Adds a method that runs upon **initialization** of a npc_interaction ( so the interactionspot within the overworld that denotes the npc interaction ) and then **loads/sets** the corresponding sprite for displaying said animal. 

Furthermore a new **exported variable** was defined that can be set to be **an invariante**, so  a special texture for a given npc. 

The method considers this invariant and prioritizes it over the default animal-sprites.

 

## Testing

The overworld now contains all the npc interactions with their corresponding sprites ( **not the final ones however** ):
image

Furthermore in the lower corner an invariant was added ( Esther is a failed square now):
image

### DIFF
--- a/Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
+++ b/Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
@@ -113,13 +113,13 @@ func set_npc_sprite():
 	var path_to_sprite:String 
 	match npc_animal_type:
 		Animal.AnimalType.SNAKE: 
-			path_to_sprite = "res://assets/art/objects/bridge_snake.png"
+			path_to_sprite = "res://assets/art/characters/snek.png"
 		Animal.AnimalType.DEER:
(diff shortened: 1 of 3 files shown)


mango-gremlin (approved):
looks really good, works well!
I especially like the name of the deer. Spider is cool too :D

PR #234, opened 2024-01-07
overworld | add methods to add/remove quests, add them to ui for tracking purpose

### LINKED ISSUE #189: overworld | update overworld UI  to accomodate active quests, animal inventory
## Motivation
As of now the Ui works only for items, however it should be able to track the following as well: 
- animal inventory 
- active quests 

### Requirements: 

1. This requires the **animaltype** representation and inventory to work well  tracked in the following issue : #190 
2. the container to track active quests ought to be contained too. represented in the following issue: #191 

## Implementation
we ought to work with signals sent from **the singleton** again and connect them to the UI upon instantiation:: 

This is the easiest solution and it works well because both nodes dont need to know each other at all!

### LINKED ISSUE #191: overworld | way to track and store quests
## Motivation
We have quests within our game that ought to be tracked: 
-> to display them 
-> to enable interaction with npcs accordingly 

We know that each quest is bound to a specific **NPC** thus we can simply use their **ID to denote the id for a quest**. 
Each Quest has the following traits which may need to be **encapsuled**? 
- issuer (npc_id)  == quest_id
- QuestType ( npc, item, bridge ) 
- target (either npc_id, itemType,bridgeEdge)

## Implementation

1. requires container ( array / list / dict ) in singleton to store each active quest
2. methods required within singleton: `add_quest` `search_quest` `remove_quest`
3. **representation** of quest either globally or within each **NPC** itself --> as they will be converted to a class anyway!

ScatteredDrifter:
## Motivation
This pr aims to add a functionality to **visually** track Quests occurring in the overworld. 
Its done by adding a new dictionary within the singleton which will track / contain each quest as string ( linked to its npc_id ) 
resolves #191 
resolves #189 
## What was changed/added

-  added **Dictionary** `quest_string_dict` in singleton_player. 
- Its modified by `add_quest_string()` and `remove_quest_string()` which are adding / removing new entries 
- added signal to notify UI for quest-updates 
- modified `add_npc_talked_to` to accommodate adding / removing a quest upon interaction
- added new method to visualize quest in `overworldui` ( similar to the previous methods, primarily as dummy until UI was added ) 

## Testing

Ran it in the overworld with its default npcs: 
before submitting "Evillyn"s quest ( talking to esther) 
image

after talking to "Evillyn" with submitted quest" 
image

so far I have not noticed any issues with this approach.

### DIFF
--- a/Arch-enemies/overworld/scripts/Npc.gd
+++ b/Arch-enemies/overworld/scripts/Npc.gd
@@ -98,6 +98,12 @@ func obtain_reward_type() -> NPC_interaction.QuestReward:
 func obtain_required_item() -> Item.ItemType:
 	return required_item
 
+# returns whether this npc holds a quest (!= NONE) 
+# or not 
+func has_quest() -> bool:
+	if quest_type != Quest.NONE:
+		return true
+	return false
 
 # --- / 
 # -- / quest behavior / interaction 
@@ -143,7 +149,9 @@ func set_quest_resolved():
 	quest_resolved = true 
 
 # checks whether conditions for completing quest were acquired 
-# updates "quest_resolved" accordingly
+# returns this state as boolean 
+# true if condition met
+# false otherwise
 func check_quest_condition() -> bool:
 	# FIXME IMPROVE WRITING / CONDITIONS 
 	if not quest_resolved : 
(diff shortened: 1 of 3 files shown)


mango-gremlin (approved):
works well and code makes sense & is structured very well!!

PR #235, opened 2024-01-07
auto assign hitboxes for water and hazards

### LINKED ISSUE #230: bridges | automatically add Area2Ds for water and hazards
## Motivation
Right now we have to instantiate the area2d for hazard and waters manually, which is a bit annoying when building levels. In line with previous issues, we can improve this by adding them automatically to our tilemap. 

## Implementation
Write a function that instantiates a given scene at a particular position for each square that is marked as WATER, SHALLOW or HAZARD, then call for them for their respective arrays. Signal handling could be an issue.

marcel-iv:
## Motivation
closes #230 

## What was changed/added
During `_ready` we collect a list of all tiles considered water, shallow or hazard. Then, we instantiate their respective scenes, add it to the grid as children, and places them according to their position within the grid. Then, add a connection from the signal `on_body_entered` to the function `on_contact_danger` which in turn simply calls the reset function for the fox. 

## Testing
Collisions are added as expected on top of their respective tiles. Note that the hitbox for the water overlaps slightly, this is required, otherwise the fox can walk on water. 
image

As you can see the fox is reset as before

https://github.com/mango-gremlin/arch-enemies/assets/104799849/31020189-c50d-42c9-8f24-d98889367861

## Additional Information
The Area2Ds and hitboxes for water and hazard are basically identical (size varies slightly, but as long as its at least 11x11, it doesn't matter). I'm not quite sure if we need them for anything else, we can delete either (and change the code accordingly) if no one has an idea why we might need it.


Kabooni:
I don't think we need the water and hazard Area2Ds and hitboxes for anything else, all they ever did was detect collision. So in my opinion, we could remove it.

Kabooni (approved):
Lgtm

marcel-iv:
I unified the two scenes and deleted the unused ones. I'll keep the PR open for a few more hours, just in case someone finds bugs

PR #238, opened 2024-01-07
bridges | Incomplete Implementation of Rotation

### LINKED ISSUE #198: bridges | implement rotation
## Motivation
Rotation adds some additional game-play depth and makes building more interesting.

## Implementation
Within the implementation of the update_grid and is_allowed either a specific rotation mode has to be added or all animals have to be internally represented as 2x2 matrices which can be easily rotated.

koennixer:
**IMPORTANT: This is incomplete and will be updated further**

## Motivation
closes #198

Rotation is an essential feature for the game.

## How does it work/Problems

Upon pressing the rotation button ("R") the animal sprite in the drag is rotated clockwise at 90°. Godot does not have an internal way of elegantly doing this. The rotation of T-Rect does not properly work for some sprites, while it does for others. If the sprite is originally square shaped (like the deer) using the internal rotation does weird things that cannot be easily fixed with an offset and reflection, while for rectangular sprite (like the squirrel and snake) the rotation is necessary. This leads to complications that have been mostly fixed, in effect much has to be hard coded and experimented with.

is_allowed obviously also needed major updating. This essentially just boils down to properly adjusting for matrix operations, i.e. checking the correct squares. This works but is most likely not bug free.

update_grid is the major problem and does not work properly. The tiles have to have alternative tiles as to be correctly displayed. This is tedious leg work with much testing. Further implementing the correct orientation for all the tiles is hard, especially because the orientation should change what tiles became ALLOWED, SIDE etc. This is honestly such a mess that it might be worth exploring just implementing a grid for each animals that just displays where goes what and then implementing matrix operations in godot. This is not something I will not do until Tuesday, but it might be necessary in the future.

## What works
Displaying the rotation works:

https://github.com/mango-gremlin/arch-enemies/assets/65815628/4f46b1ba-31bd-4868-8d09-73efa7a4966a

The game also checks mostly correctly if the rotated body is allowed to be placed. (This cannot be elegantly shown with a video because the cursor doesn't always update nicely, just believe me)

Some placement for rotated animals work:

https://github.com/mango-gremlin/arch-enemies/assets/65815628/6b03e702-be72-431b-8a24-8a1d179cda6f

## What doesn't work
For some animals the rotated placement has not been implement or does not work properly. This has to be worked one, but the code is just very tedious.

## Additional Information
This is the biggest feature effort wise (in my opinion). Until Tuesday a functioning demonstration should work but I cannot promise it will be bug free. Further updates will be made but given how complex this one is because of the amount of hard-coding an unfinished update seems helpful for reviewing.


mango-gremlin:
Since we are planning not to implement this at all, should we close this pr, but keep the branch for now?

PR #239, opened 2024-01-07
overworld | automatically entering bridge-level denoted by BridgeEdge

### LINKED ISSUE #225: overworld | functionality for entering a given scene
## Motivation
as of now entering a scene ( bridge-level mostly ) was done by switching to that scene statically.
**With the rise** of more levels we want this to be dynamic -> taking the correct level-scene based on the given bridge-Edge! 

## Implementation
a method that takes **a bridgeEdge** and matches against its values to select a **string** that will denote the level linked to it. 

**the level is then** opened accordingly

ScatteredDrifter:
## Motivation
resolves #225 
This PR adds functionality to enter a bridge-level denoted by its linked ( desired ) BridgeEdge  ( so the connection from one island to another (i.e (0,1) ). 
This is done by mapping a connection of a bridge to a "scene" which is referenced by a string. 
A method 'obtain_bridge_scene' then sets a given path if available and returns it. 

The player can then load this scene or not. 

## What was changed/added
- added Dictionary mapping bridge-level to a scene in 'singleton_player'
- incorporate method 'obtain_bridge_scene' to retrieve a path 
- add methods to BridgeEdge class for comparing them against each other 'BridgeEdge.is_same_as(bridge_edge_to_compare)'
- update 'enter_bridge_scene' in **Player** which now enters the scene after requesting its source. 
- update **main-overworld** to accommodate the new mapping
- added further samples to **overworld_test_area**, especially to test whether '(0,2) == (2,0)' when interacting and deriving a level

## Testing

Code is likely not the most beautiful yet and could be improved I think. **please give feedback on that!**

yet it works as intended: 

**Start the overworld_test_area** scene and try 
- interacting with a solved bridge -> 0,1 and back 
- interacting with unsolved ( but available!) bridge_connection: (0,2) and (2,0)
- interacting with and unsolved and unavailable  bridge_part (0,3)


mango-gremlin (approved):
Honestly, this implementation works well and in my opinion is very well thought out. I think the code is also completely fine, and the problems that presented themselves were solved in a sensible way. Since we are only trying to have the working game at the moment, I would say this is completely fine!

PR #241, opened 2024-01-07
Issue172/bridge inventory

### LINKED ISSUE #172: bridges | Implement basic animal Inventory
## Motivation
Way to track which animals you have in your party/and can use for bridges.

## Implementation
The implementation should provide a basic inventory:

This inventory should be like the inventory already implemented in the overworld.

It should have the following functionality
- saves how many of each animal you have (should be expandable to work with the singleton in the future -> so that it can take the inventory content information from the player in the future)
  The amounts of each animal should/could be saved in a dictionary where key = AnimalType (as represented in the singleton) and value = how many you have 
  It would be great if each value, before it is saved into the dict, is exported with @export to that you can edit the animal amounts in the godot inspector.
- save how many animals have been used so far
- display and disable placement if animal not available
- the inventory should only take info out of the singleton, *not edit it* unless the level is completed and the player goes back to the overworld.

There should be a basic UI as follows:
- hbox container that contains buttons for each animal type & shows how many you have of each type
- the buttons should spawn the respective animal
- spawning animal should update inventory

Raining-Cloud:
## Motivation
closes #172 

Implements a basic inventory behavior with animals for the bridge building linked to the global inventory provided by @ScatteredDrifter.
Also found a bug which caused the grid.tscn Node getting uneccesarry duplicated.

## What was changed/added
Added new variables to the grid. for the revert option. And a new condition in the grid_drag.gd checking if we have enough animals

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/59502349/5c718afe-ad1f-4e25-9993-8e57e747d469

If you wan't to test this locally I suggest adding this code into the `_ready()` function in grid.gd

```
SingletonPlayer.add_to_animal_inventory(Animal.AnimalType.DEER, 2)
SingletonPlayer.add_to_animal_inventory(Animal.AnimalType.SNAKE, 2)
	
start_animals = SingletonPlayer.get_animal_inventory().duplicate(true)
```

## Additional Information
We might need to be careful if we exit the bridge building, due to currently we do not reset the animal inventory count!

### DIFF
--- a/Arch-enemies/bridges/script/grid.gd
+++ b/Arch-enemies/bridges/script/grid.gd
@@ -10,6 +10,11 @@ var tile_size = tile_set.tile_size
 @export var y_size = 22
 @export var square_size = 10
 
+@export var inventory_deer : Label
+@export var inventory_snake : Label
+@export var inventory_spider : Label
+@export var inventory_squirrel : Label
+
 #And finally some values we need later
 var grid_size = Vector2(x_size, y_size)
 var grid = []
@@ -17,6 +22,10 @@ var start_grid = [[[]]]
 var last_states = []
 var save_states = 10
 var state = 0
+
+#for the inventory
+@onready var start_animals : Dictionary = SingletonPlayer.get_animal_inventory().duplicate(true)
+var placed_animals : Array = []
 
 # ids of all tilemap layers
 const BACKGROUND_LAYER_ID = 0
@@ -51,9 +60,13 @@ signal current_grid(current_grid)
 enum ENTITY_TYPES {GROUND, WATER, AIR, ANIMAL, FORBIDDEN, ALLOWED, SIDE, BOTTOM, SHALLOW}
 
 func _ready():
+	#update the ui
+	update_inventory()
+	
 	#We save the previous states of the grid in an array, this array is initalized here
 	for i in range(save_states):
 		last_states.append([[]])
+		placed_animals.append(Animal.AnimalType.NONE)
 	#Here we iterated over the grid and fill it
 	for x in range(grid_size.x):
 		grid.append([])
@@ -183,6 +196,10 @@ func update_grid(pos, data):
 	var animal = data["animal"]
 	var x = pos.x
 	var y = pos.y
+	
+	#Whenever we change the grid, i.e. update it, we have to track that here
+	state += 1
+	
 	#Based on the animal we use the apprioate filling
 	match animal:
 		"DEER":
@@ -211,6 +228,10 @@ func update_grid(pos, data):
 			for position in new_bottom:
 				if(grid[x + position.x][y - position.y] == ENTITY_TYPES.AIR):
 					grid[x + position.x][y - position.y] = ENTITY_TYPES.BOTTOM
+					
+			#this seems to be the only way to keep track of the placed animals
+			placed_animals[state % save_states] = Animal.AnimalType.DEER
+			SingletonPlayer.add_to_animal_inventory(Animal.AnimalType.DEER, -1)
 		"SNAKE":
 			#Snake works just like Deer
 			var new_allowed = [Vector2i(0, 1), Vector2i(1, 1), Vector2i(2, 1), Vector2i(3, 1)]
@@ -235,6 +256,10 @@ func update_grid(pos, data):
 			for position in new_bottom:
 				if(grid[x + position.x][y - position.y] == ENTITY_TYPES.AIR):
 					grid[x + position.x][y - position.y] = ENTITY_TYPES.BOTTOM
+			
+			#this seems to be the only way to keep track of the placed animals
+			placed_animals[state % save_states] = Animal.AnimalType.SNAKE
+			SingletonPlayer.add_to_animal_inventory(Animal.AnimalType.SNAKE, -1)
 		"SPIDER":
 			#Spider is the easiest, nothing much happens here
 			var new_allowed = [Vector2i(0, 1), Vector2i(1, 0), Vector2i(0, -1), Vector2i(-1, 0)]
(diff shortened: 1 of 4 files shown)


Arch-enemies/bridges/script/grid.gd
```diff
@@ -211,6 +227,10 @@ func update_grid(pos, data):
 			for position in new_bottom:
 				if(grid[x + position.x][y - position.y] == ENTITY_TYPES.AIR):
 					grid[x + position.x][y - position.y] = ENTITY_TYPES.BOTTOM
+					
+			#this seems to be the only way to keep track of the placed animals
+			placed_animals.append(Animal.AnimalType.DEER)
+			SingletonPlayer.add_to_animal_inventory(Animal.AnimalType.DEER, -1)
```
ScatteredDrifter, line 233: wouldn't it be more beneficial if this update only occurs whenever the scene is left? 

Meaning that whenever you leave the current bridge scene it propagates its inventory back to the singleton ( and update the amount of animals available based on the local inventory you copied before)

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -90,10 +90,10 @@ func use_item(requested_item:Item.ItemType):
 # --- / 
 # -- / animal inventory
 
-func add_to_animal_inventory(new_animal:Animal.AnimalType): 
+func add_to_animal_inventory(new_animal:Animal.AnimalType, quantity:int = 1):
```
ScatteredDrifter, line 93: reasonable change, thanks :)

Arch-enemies/bridges/script/grid.gd
Kabooni, line 233: All this repetitive code should be refactored into a function, like e.g. remove_from_inventory(Animal.AnimalType)

Kabooni (changes requested):
From what I've seen, it works great! 
But having another container for the inventory count is bloating our main node tree, is it possible to combine all inventory-related stuff into one separate scene? Or at least combine as much as possible?
And it would be nice if you could improve the visual design a bit more. In detail, the animal counts should align with their respective animals, and the whole inventory should be at the bottom. In the end, we want to achieve something like this:
grafik
Of course, you don't have to draw anything, but everything should be in the right place to make it easier for the future. If it looks completely terrible for now or there isn't enough space, just let me know.

Arch-enemies/bridges/script/grid.gd
Raining-Cloud: Yeah it would be better, but the only solution for this would be to intriduce another singleton which handles the brige inventory which I tried to avoid

Kabooni: Part of the issue description was to find a way to handle everything _without_ having to edit the player singleton all the time, and if you need another singleton for that, it's fine.
I mean you could just use grid.gd or the animal count labels as a "local" singleton and create an inventory dictionary as suggested in the issue. When clicking the return button in the goal menu, you could signal that to grid.gd and change the player singleton based on the inventory.

Kabooni: Even repositioning the inventory and aligning the labels?

Kabooni: Additionally, it would make it much easier to export the inventory. We need an easier way to change the inventory values for testing purposes, rather than having to change the code.

Raining-Cloud: Ok it seems that you can move the animals, this is implemented statically so I will give it a shot. the align is probably easy, I just need to check how GoDot UI works

Kabooni: Or in worst case scenario remove just the label container and align the labels separately

ScatteredDrifter: > and if you need another singleton for that, it's fine. 

Tbh I disagree, A singletone would be loaded constantly and persist even if the bridge-scene ( and therefore its required context ) is left / terminated

koennixer (approved):
Many of the issues raised above are valid. What I would like to caution is that this change is behind the current main significantly. Many of the previous bugs and code smells are still here. The inventory it self, for just its own works.

The issues currently being discussed, as far as I can tell, is how to handle singleton values and also the positioning and appearance of the animals. These are, in my mind, completely separate issues that can be solved separate of just this inventory issue. Specifically making the position of the sprites not hard coded is something that can solved it's own issues and is a major point of contention here. I would prefer it if we got around to merging this soon and then address the other problems in their own issues to avoid this code falling even further behind. Any remaining oddities, like the handling of singletons, can be fixed in a refactoring at the end when we know the full scope of what we need to handle globally.

mango-gremlin (changes requested):
So I agree with all three of you here on separate things
- we agreed on not refactoring for now, so this should be left out of this pr. However it would be nice if you tried to write code that doesnt need to be refactored in the first place, as code quality is important.
- positioning of animals was not something that was part of the issue. So there should be a separate issue for this
- HOWEVER: currently the singleton is being edited the entire time, as far as I understand? This is specifically something that the issue mentioned not to do. I did say this was okay, so thats my mistake. But i just thought about it again and what happens if you leave the bridge level? Then it doesnt save your animals in the bridge level (this is not a feature that will be included in the final game) - so you just lose animals? this shouldnt happen. Of course you could just reset everything - performance wise it shouldnt matter- but thats really smelly. So you should save the amount of each animal in a separate data structure. And edit this separate data structure. Only interact with the singleton when querying the animal amounts at the beginning, and when the level is solved to update the animal amounts.

Arch-enemies/bridges/script/grid.gd
ScatteredDrifter: > Ok it seems that you can move the animals, this is implemented statically so I will give it a shot. the align is probably easy, I just need to check how GoDot UI works

I would argue that this implementation does not have to implement the UI-elements correctly. 
-> Primarily because this is mostly a design choice and should therefore be handled by the ui/assets team ? 

Of course, to implement the UI it requires the base functionality ( i.e having an inventory that updates accordingly and can be queried ) however I would say that the actual implementation of the ui should be resolved in a separate branch or at least not solely by the person doing the back-end. In other words the ui-assets might be required to take a look and add it too ( at best together in person or remotely ^^)

Kabooni (approved):
Alright I see all your points. The code works of course, so after @mango-gremlin's request was implemented we can merge it and do everything else later.

ScatteredDrifter:
Changes to convert to a local inventory were incorporated in #262 .
The branch should be used as it contains newer changes, is up to date with main and satisfies the requirements. 

Closing this PR is advised, as its superseded by #262

ScatteredDrifter:
**\\** closed as #262  provides the more up to date version of this feature **//**

PR #242, opened 2024-01-08
bridges | proper placement of animals

### LINKED ISSUE #221: bridges | animals can't always be placed properly
## Description
Because placement of animals is based on the mouse at the lower left corner of an animal you cannot place it even when you should be able to.

**To Reproduce**
Try to place deer underneath of spider.
Or try to place a snake somewhere with its body on top of something but its tail in the air.

**Expected behavior**
Placement shouldn't be only allowed where your mouse is but also at all other tiles of an animal.

**Screenshots**

https://github.com/mango-gremlin/arch-enemies/assets/116217918/553cf871-60b0-40ed-8651-59bd721f431d

## Solution Proposal
Base placement not only on tile at mouse cursor but also all other tiles of the animal. Iterate through range("size) and see if any tiles are in allowed places.

aylinfi:
## Motivation
closes #221 

Animals could not be placed everywhere they were allowed.

## What was changed/added
When looking if a placement is allowed we now iterate through all tiles of the animal to see if it can be placed.

## Testing

grafik

## Additional Information
Placement is still based on the lower left corner, which can feel a bit awkward. But I couldn't figure out how to change that.


Arch-enemies/bridges/script/grid_drag.gd
```diff
@@ -106,65 +106,62 @@ func _drop_data(at_position, data):
...
-	#This works just like Snake
-	var is_allowed = false
-	if(grid[pos.x][pos.y] == ENTITY_TYPES.ALLOWED 
-	or grid[pos.x][pos.y] == ENTITY_TYPES.SIDE or grid[pos.x][pos.y] == ENTITY_TYPES.BOTTOM):
-		is_allowed = true
-	return is_allowed
-
+	var pos_Type = grid[pos.x][pos.y]
+	var is_allowed_and_free = pos_Type == ENTITY_TYPES.ALLOWED \
+		or pos_Type == ENTITY_TYPES.SIDE or pos_Type == ENTITY_TYPES.BOTTOM
```
koennixer, line 134: This kind of code is extremly unreadable. It does not hurt to extend this to more than two lines to make it more legible.

koennixer (approved):
This works. Nothing much to say, it is functional an bug free as far as I can see. I will merge it  immediately.

PR #244, opened 2024-01-08
bridges | Fixed boundary issues

### LINKED ISSUE #214: bridges | dragging snake head outside grid causes crash
## Description
Current implementation of the snake's head while dragging causes a crash when moving it past the screen.

**To Reproduce**
1. Pick up the snaek
2. Drag it outside the right boundaries of the grid
3. Game crashes (invalid index at array, in ``grid_drag.gd`` (line 121 currently, while checking the snake's head)

**Expected behavior**
Player is simply unable to correctly place snake.

## Solution Proposal
Check if position is within grid boundaries, before acting on it. This *might* be an issue in other parts of the code, please check if the problem could occur elsewhere (for example, placing a snake close to the right end, so that part of it extends beyond the grid might also cause a crash).

koennixer:
## Motivation
closes #214

When dragging the cursor to the edge it could crash the game because wrong array elements might be accessed.

## What was changed/added
An out_of_bounds function was added that checks if the current cursor position would check an illegal square for the current
animal. This is based on the animals sprite size (which is very handy to implement for other animals).

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/65815628/c3b5a870-7528-42f1-a6ad-6a26e4834f1b

### DIFF
--- a/Arch-enemies/bridges/script/grid_drag.gd
+++ b/Arch-enemies/bridges/script/grid_drag.gd
@@ -153,7 +153,7 @@ func is_deer_allowed(pos):
 			if(Vector2i(delta,epsilon) not in ignore):
 				var pos_Type = grid[pos.x+delta][pos.y-epsilon]
 				#Is there something to attach the animal to?
-				if(pos_Type == ENTITY_TYPES.ALLOWED):
+				if(pos_Type == ENTITY_TYPES.ALLOWED or pos_Type == ENTITY_TYPES.SHALLOW):
 					is_allowed = true
 				#Is there space for the animal?	
 				if(pos_Type == ENTITY_TYPES.FORBIDDEN or pos_Type == ENTITY_TYPES.GROUND or 
--- a/Arch-enemies/bridges/script/grid_drag.gd
+++ b/Arch-enemies/bridges/script/grid_drag.gd
@@ -108,7 +108,7 @@ func _on_grid_current_grid(current_grid):
 	grid = current_grid
 	
 func is_snake_allowed(pos):
-	if(out_of_bound(pos, 5, 1)):
+	if(in_bound(pos, 5, 1)):
 		return false
 	var is_allowed = false
 	var is_free = true
@@ -131,7 +131,7 @@ func is_snake_allowed(pos):
 	return is_allowed and is_free and head_free
 
 func is_spider_allowed(pos):
-	if(out_of_bound(pos, 1, 1)):
+	if(in_bound(pos, 1, 1)):
 		return false
 	if(pos.x > grid.size() or pos.y > grid[0].size()):
 		return false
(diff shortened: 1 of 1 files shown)


Kabooni (changes requested):
Now I can't place anything anymore. I'll update the comment after I find the cause, should I be faster than you.
Update: Something is very wrong with out_of_bound(), it's basically always true.

koennixer:
I made it actually work now. Sorry, should have tested more extensively.

marcel-iv (changes requested):
I still found some minor bugs:
- placing an animal in the leftmost square is impossible
- placing an animal in the rightmost square is impossible

Another thing to consider is that sometimes, you might want to place an animal slightly outside the screen (lets say the leftmost part of the animal would be outside the grid). I think its more intuitive to allow these types of placement. However, this is relevant to a different part of the code and probably unreasonable to do until tomorrow. 

We can merge once the bugs above have been adressed.

Kabooni (approved):
It works now, but you can't drag a deer into shallow water anymore. But this bug is known and will be addressed elsewhere.

Kabooni:
> I still found some minor bugs:
> 
>     * placing an animal in the leftmost square is impossible
> 
>     * placing an animal in the rightmost square is impossible
> 
> 
> Another thing to consider is that sometimes, you might want to place an animal slightly outside the screen (lets say the leftmost part of the animal would be outside the grid). I think its more intuitive to allow these types of placement. However, this is relevant to a different part of the code and probably unreasonable to do until tomorrow.
> 
> We can merge once the bugs above have been adressed.

We could just increase the grid size by two squares, but leave the viewport size as it is. Then it should theoretically be possible to place an animal slightly "outside" the window, but it would also make it possible to place e.g. a spider completely outside. It's not a perfect fix, but I think enough for tomorrow.

koennixer:
> I still found some minor bugs:
> 
> * placing an animal in the leftmost square is impossible
> * placing an animal in the rightmost square is impossible

This is more a simplification that it is a bug. Most animals have square around it with tiles that are to be allowed or forbidden, i.e. the snakes head. If you account for all these possibilities it makes the function extremely bloated and much less intuitive. Personally I feel like the ability to place in these squares, which will be hardly relevant, is not really worth it.

marcel-iv:
> We could just increase the grid size by two squares, but leave the viewport size as it is. Then it should theoretically be possible to place an animal slightly "outside" the window, but it would also make it possible to place e.g. a spider completely outside. It's not a perfect fix, but I think enough for tomorrow.

Thats actually a really good idea, although a bit hacky for my taste. Should be enough for the presentation, although I think the current implementation is sufficient already. Additionally, we should really think about getting the grid size automatically, the current solution isn't great.

marcel-iv (approved):
In that case, we can merge now, and implement a slightly more dynamic system that allows placement on the sides of the grid on a later date

PR #245, opened 2024-01-08
bridges | implement hierarchy for tiles

### LINKED ISSUE #219: bridges | implement hierarchy for tiles

### LINKED ISSUE #231: bridges | update some placement rules
## Motivation
I noticed that in some niche cases, animal placement doesn't behave like expected. These include:
- after placing an deer in water, you can place squirrels sideways to it in water
- when placing a spider above water, and a `SIDE` shore zone is on top of it, it doesn't get updated as `ALLOWED`, despite the spiders ability and it being a top dropzone
- `SIDE` and `BOTTOM` *might* block extended placement of some animals. This was an issue briefly in #218, but has been fixed there. Not that this is right now not an issue with any existing animal (as far as I'm aware at least), due to our animal sizes and placement logic, but updating existing rules to account for any change in the system might be a good idea.

## Implementation
Update relevant code in the allowed placement functions of animals. 

## Additional Notes
Might be a good idea to do this after #221  and #219 , which might fix some of these issues.

koennixer:
## Motivation
closes #219 
closes #231 (at least I think it should fix it)

## What was changed/added
When changing a tile in update grid it now correctly checks the hierarchy using tile_update.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/65815628/508177e6-f6f7-4ba4-aef5-ca4651784be0

### DIFF
--- a/Arch-enemies/bridges/script/grid.gd
+++ b/Arch-enemies/bridges/script/grid.gd
@@ -347,15 +347,17 @@ func last_state():
 			color_grid()
 
 func tile_update(pos, current, next):
-	if grid[pos.x][pos.y] == ENTITY_TYPES.AIR:
+	#We check what the current square is, based on what we want to put there next we either do it
+	#Or do not do it. Hierarch is currently FORBIDDEN > SHALLOW > ALLOWED > SIDE > BOTTOM > AIR
+	if current == ENTITY_TYPES.AIR:
 		grid[pos.x][pos.y] = next
-	elif grid[pos.x][pos.y] == ENTITY_TYPES.SHALLOW and next == ENTITY_TYPES.FORBIDDEN:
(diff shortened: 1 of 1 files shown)


marcel-iv (approved):
looks good, very nice that everything is done in a single function. 

I'm a bit confused as to why `current` is specified as a parameter and never used, maybe you could change the code to include it, and make it a bit more readable, maybe also add a brief comment, even though the code is self explanatory. If theres no time for that, just merge it

koennixer:
This is all very true. I have made it less terrible and remembered to actually use current. I will merge it now.

PR #255, opened 2024-01-12
bridges | Added the dynamic grid

### LINKED ISSUE #248: bridges | automate grid size
## Motivation
The grid size is hardcoded and should be computed dynamically. 

## Implementation
how should this be implemented?

koennixer:
## Motivation
closes #248

To implement additional levels it would be good if the grid adjusts automatically so we can just repeatedly implement the same scene.

## What was changed/added
Camera has it's own script now. It just sends it current zoom. Since camera is often used for many things giving it it's own script is future proofing. On ready the grid.gd now signals to get the zoom and with that adjusts the grid size.

## Testing
everything is peaceful

The grid looks the same even though the sizes are not hard coded.

## Additional Information
Note that if you adjust the zoom you have to adjust the camera position, but this is trivially true.

### DIFF
--- a/Arch-enemies/bridges/scenes/Grid.tscn
+++ b/Arch-enemies/bridges/scenes/Grid.tscn
@@ -1,6 +1,7 @@
-[gd_scene load_steps=33 format=3 uid="uid://bl1jporj7hosu"]
+[gd_scene load_steps=34 format=3 uid="uid://bl1jporj7hosu"]
 
 [ext_resource type="Texture2D" uid="uid://d0o2q6fijmd1" path="res://assets/art/tilesets/bridge_tileset/bridge_background_color.png" id="1_uowbf"]
+[ext_resource type="Script" path="res://bridges/script/Camera.gd" id="1_vkapa"]
 [ext_resource type="Script" path="res://bridges/script/grid.gd" id="1_xtvvi"]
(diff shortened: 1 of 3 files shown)


Kabooni (approved):
I'm not 100% sure how the code works, but I didn't encounter any bugs.
If another person is also fine with it, you can merge it.

marcel-iv (changes requested):
While this doesn't seem to cause any bugs whatsoever, I'm also not quite sure it works like intended. When placing tiles beyond the camera, the grid size stays static at (115, 64), which is much more than expected, and doesn't update. Since we maybe want to build longer levels that exceed camera boundaries, this can become an issue in longer levels. 

However, it still works as expected, and as long as our tilemap doesn't exceed the boundaries of the current size I suppose. I leave it up to you if you want to invest more time into this, but since its unlikely to cause any issues in the part of the game we want to show, we can also just note that it doesn't behave as expected, and should be updated for very large levels. We probably have more pressing issues to invest our time in right now. 

Let me know how you want to proceed, and I can approve this PR, as it doesn't break anything, as far as I've noticed, and works for our levels.

koennixer:
I have fixed the issue with the grid size. I forgot that we store the grid size in a vector later.

If you want levels of a size larger than the screen then you have to tell what that size is based on. The only thing I can think of is some sort of singleton or global value that is retrieved on ready, but that requires a mapping of level_id to grid_size. None of this has been defined thus far and is not part of this issue. So the window size is the only thing to go off of.

marcel-iv (approved):
thanks for fixing it so quickly, we can merge this PR now

PR #259, opened 2024-01-13
Overworld | Simplify Building a bridge in editor

### LINKED ISSUE #184: overworld | constructing bridges should be easier
As noted in the wiki its a little tedious to set / define a given bridge and its connections. 

Especially it requires users to **add new nodes** that need a specific name --> because they cannot be accessed otherwise. 
They are used for: 
- grouping visual bridge-parts that can either be hidden ( unbuilt ) or seen ( built ) 
- setting the collision box which blocks the path when the bridge is unbuilt. 

This should be improved, maybe by using a template-like approach, which creates an extra script that binds two new bridge_interactions together and then only requires the given bridge_parts for it too? 

Below you can see the current structure that needs to be established for **each new bridge connection** because we cannot include it in the node itself. 
image
it results with the following: 
image

---

**Why** we cannot fill the node itself: 
Afaik a node which is then instantiated is **static** in its internal components meaning that we cannot adjust the amount of bridge-parts or the position / size of the collision afterwards. 
--> This means that we have to somehow set those specific nodes as child after we've instantiated the base node

ScatteredDrifter:
## Motivation
resolves #184 

This PR aims to make **constructing** a bridge in the actual overworld easier.
What it means: 
- placing a new bridge connection with start/end should be as easy as dropping a template
- increasing length  of bridge-connection (sprites) should be easy
- aligning bridge_interaction with overworld should be easy 

**IMPORTANT**:
This still requires **new assets** that will be provided by @mango-gremlin .
**After those were added** --> merge / close this PR!

## What was changed/added

2 new templates (scenes) were added:
- `horizontal_bridge_template`
- `vertical_bridge_template`
Those contain the following:
- sprites that are displayed **upon the level was solved**
- 2x bridge_interactions scenes --> to interact with either side 
- **collision** to disallow walking there **if it is unsolved**
- `horizontal/vertical bridge_part_with_collision` scenes that will **contain the assets** for the bridge-parts that connect to starting shores.

visualized those look like this:
image

## Usage | will be merged to WIKI:
Creating a new bridge requires the following: 
1. Updating the tilemap
2. Adding a new template and adjusting it 

**Tilemap**: 
Because the tilemap manages collision to disallow leaving an island we require "special shores" that  are place wherever one will place a bridge-connection.
Those are found  in the "alternative-tiles" section of the **tilemap** and dont have any collision ( remember those collision are provided by the template!) 

Once those were added we also require either a **horizontal/vertical** starting-point for the bridge ( sprite indicating a bridge-start)

**overworld-scene**:
After the **tilemap** has been updated we have to add the bridge-template to provide its functionality ( and to prevent the player from leaving the island into water ). 
For that: 
1. add either **horizontal/vertical** bridge template
2. **right-click** this node and mark checkbox "editable children" 
3. you can now move this templates content individually and align it with the tilemap ( this includes: **starting point**, position of **bridge-parts**. 
4. you can **extend/shorten** the bridge by adding new `vertical/horizontal_bridge_part_with_collision` as children of the `node` `bridge_parts` --> those are hidden if the bridge is unsolved! 
5. You have to set the "bridgestart/dest-ids" in the **bridge_interaction** for each site ( so that they know which edge they are in our graph representation )
6. test your changes! 

## Testing
Gave a small showcase in meeting with @mango-gremlin , they will provide the assets required.
Works well otherwise.


mango-gremlin:
added the sprites for the bridge templates and adjusted the collision as well as set the z-index of bridge elements to -1 -> player can now walk on the bridge, not below it. Was the easiest way to do this, might be adjusted later with separate sprites for snake and for bridge floor.
image


https://github.com/mango-gremlin/arch-enemies/assets/104830903/9b1cd041-63c8-425f-b654-a869a445e15b

ScatteredDrifter:
nice, looking good :)

mango-gremlin (approved):
this works well and is a good way to generate bridges quickly.

ScatteredDrifter:
@mango-gremlin the assets ( or at least some placeholder ) for the bridge-starts for **the tilemap** are missing and should be added. 

I would argue that once those have been added, we **should give them the collision** instead of the round collision-shapes defined in both templates right now.

ScatteredDrifter:
@mango-gremlin  took a look at your most recent change, thanks for the work / contribution. 

I would approve to merge now //

PR #261, opened 2024-01-13
bridges | Make the background more distinguishable from the foreground.

### LINKED ISSUE #260: bridges | Background Improvement (darker)
## Motivation

Better visual clarity is good for the game! The player wants to see what they can interact with and what is part of the background.

## Implementation

New art thats darker in order to contrast with the foreground.

Raffael-Rack:
## Motivation
closes #260

The mechanically/game-play relevant parts of the foreground didn't stand out enough from the background.

## What was changed/added

I changed the background sprites to a dark forest and added a layer of extra shade in the middle. Should we change it further?

## Testing
This is the background:

New_Background_3


mango-gremlin (changes requested):
This looks really good!
Although while testing it I still thought it is a bit too much with the many trees. 
-I think in the bridge building area there shouldnt be too much going on, since the animals are quite detailed and colourful. Could you lower the opacity of the blass grün trees? And in general, maybe use more similar colours for the different trees, so that there is visibly something there, but not too many colours/contrasting objects. 
- Furthermore, depending on my screen i could not see a difference with the darker layer. I would increase the opacity of the darker layer AND: i wouldnt make it black, but green, and the green shouldnt be to close to black, i think its dark enough with that.

Raffael-Rack:
New_Background_4

I simplified the Background by removing details to further distinguish it from the foreground.

mango-gremlin (approved):
thats much better, its now much less distracting, thank you!

PR #262, opened 2024-01-13
bridges | bridge inventory but local first

### LINKED ISSUE #172: bridges | Implement basic animal Inventory

ScatteredDrifter:
## Motivation
this resolves #172 like #241 does but **with a local implementation**

## What was changed/added

We load the singleton animal-inventory once and then use it locally until a function to **leave the scene** is called to "set the update animal inventory to _global state_".

I've taken @Raining-Cloud methods to implement the inventory and adapted them to comply with the _local only_ aspect. 

## Testing

You should be able to run it locally an start the "grid" scene. 
~~There you will have an inventory of 1,1,0,1 which could also be changed with the method `set_animal_inventory()` within "Grid". **This acts as placeholder for the Singleton Inventory** because it defaults to an empty inventory making it harder to test.~~ 
You can set the inventory by manually setting an inventory in the Singleton --> animal_inventory

~~**IMPORTANT**: Before merging this initialization should be updated to load from the **singleton** instead! 
Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.~~
done in efbbbff


ScatteredDrifter:
merged main to get rid of the conflicts.

aylinfi (approved):
Inventory seems to work, code looks good too. 
Are the FIXMEs supposed to stay for now?

ScatteredDrifter:
> Inventory seems to work, code looks good too. Are the FIXMEs supposed to stay for now?

most are not, I've removed them and further unused statements / expressions from this implementation. 
Could you have another look @aylinfi ?

aylinfi (approved):
lg but all the animals in the inventory are set to 0 now?

ScatteredDrifter:
> lg but all the animals in the inventory are set to 0 now?

@aylinfi 
Well yes because on default the player does not have any inventory. 
As seen in the singleton_player you can set the player-inventory with a debug-inventory to have  non 0 values.

PR #196, opened 2023-12-27
Added basic overworld dialogue system

DerSimi:
# Dialogue System
image

This PR builds on #183, it resolves #145 and #188

It is possible to create custom dialoges and add it to npcs.

# How create a custom dialogue 
Firstly, navigate to the folder `dialogues` in the overworld. Create a subfolder with a name of your choose.

Secondly, create some pages for your dialogue. They have to extend Dialogue_Entry as done in the existing example. The structure is like this:
```
extends Dialogue_Entry


class_name Page1

# --- /
# -- / class constructor 
func _init():
	pass

func content() -> String:
	return "Hallo Welt"

func image_src() -> String:
	return "res://overworld/dialogues/test_dialogue/red.png"
	
# null ement in array means here disable button X
func btn_text() -> Array[String]:
	return ["Back", "Weiter", "X"]

# btn button in from left to right, starting with 0
func btn_action(btn: int):
	pass
```
Where title represents the first line of the dialogue page, as seen in the example picture, holding a content "Hallo Welt".

btn_text is a three element array with the buttons names. They can change. To just use the first buttons and disable the others, one could write:
```
func btn_text() -> Array[String]:
	return ["Beenden", "", ""]
```

`btn_action` is called when an button is pressed. It is only called for enabled buttons (See `btn_text`). Where `btn` is the button id, with a maximum range of `0` to `2`. (The button name mentioned first in `btn_text` corrosponds to `button id = 0`)

Each page should have an own file in your dialogue folder.

Now create a dialogue data in the same folder, it has the following structure:
```
extends Dialogue_Data

class_name ExampleData

# --- /
# -- / class constructor 
func _init():
	entries = [Page1.new(), ...]
```
Register your pages there.

# What should I do in by `btn_action` method?
What you want.

Some useful calls inside the `btn_action` method:
`SingletonPlayer.exit_dialogue()` - exists the dialogue (means closing the window and reallows movement of the player)
`SingletonPlayer.dialogue.select_page(1)` - selects another dialogue page of YOUR current dialogue. The first page has index 0.
`SingletonPlayer.dialogue.finish_dialogue()` - marks the dialogue as FINISHED. It does not close the dialogue window.

# How to link my dialogue to a npc?
This is rather simple. Navigate to dialogue/Dialogue_Registry.gd in the overworld. You find a dictionary there:
`static var npc_dialogues = {1: ExampleData.new()}`
Insert your dialogue data there. The key is the npc id
# How to open the dialogue?

[...]

Issue #188, opened 2023-12-26
overworld | resetting dialogue-statue automatically after leaving dialogue


ScatteredDrifter:
## Motivation

currently _( in the changes introduced in #183 )_ we have a keypress **Q** that helps to leave the dialogue **and move again**.
This is not desired and was added to have a testing structure for the upcoming dialogue-sytem
	to be found in `Player.gd` 
	`func check_input():`

## Implementation

This functionality is part of #145  and should be considered there too

ScatteredDrifter:
has been resolved with #266

PR #266, opened 2024-01-14
Added basic dialogue system to the overworld

### LINKED ISSUE #145: Overworld | dialogue-system that can be used by NPCS
## Motivation
We need a way to handle interactions and **a conversation** with an NPC: 
- the dialogue should not restart completely every time 
- we need an **abstract** option to easily add $n$ dialogue options
- the dialogue / answers by the player should also be contained 
- an dialogue can change by selecting buttons ( like " can you help with that? " : yes : no ) 
- this decision should be saved --> considering that we are mostly using that for **quests** we can just set "has_quest" to true afterwards, saving the state that the player accepted that request

## Requirements: 
This implementation should incorporate / follow the given concepts: 

**modularity**: ( easily setting where ( in the dialogue) a decision is required or when one only proceeds **if** an item was provided ) 

**setting up easily**: feeding dialogue and options in easily ( like providing an easy way to initialize the Dialogue-structure ) --> **we may avoid** hard-coding those dialogues and instead have them be set dynamically upon creation of it

**ease of usage**: Using this whole structure should be as easy as: 
1. creating instance 
2. requesting dialogue 
3. the system handles what to **display**, **require** (like an item or interaction), and also **its state**

## Implementation

Maybe a separate class could be implemented? 

1. one could instantiate it with the amount of dialogues etc at the beginning
4. possibility to save the state of the communication ( like a number indicating position in "list of Dialogue-strings" 
5. **upon construction of the dialogue-object**:  we can also set **in which state a dialogue** should occur, and when not.

DerSimi:
# Dialogue System
image

This is not the first PR regarding this: https://github.com/mango-gremlin/arch-enemies/pull/196

It resolves #145 and #188

It is possible to create custom dialoges and add it to npcs.

# How create a custom dialogue 
Firstly, navigate to the folder `dialogues` in the overworld. Create a subfolder with a name of your choose.

Secondly, create some pages for your dialogue. They have to extend Dialogue_Entry as done in the existing example. The structure is like this:
```
extends Dialogue_Entry


class_name Page1

# --- /
# -- / class constructor 
func _init():
	pass

func content() -> String:
	return "Hallo Welt"

func image_src() -> String:
	return "res://overworld/dialogues/test_dialogue/red.png"
	
# null ement in array means here disable button X
func btn_text() -> Array[String]:
	return ["Back", "Weiter", "X"]

# btn button in from left to right, starting with 0
func btn_action(btn: int):
	pass
```
Where title represents the first line of the dialogue page, as seen in the example picture, holding a content "Hallo Welt".

btn_text is a three element array with the buttons names. They can change. To just use the first buttons and disable the others, one could write:
```
func btn_text() -> Array[String]:
	return ["Beenden", "", ""]
```

`btn_action` is called when an button is pressed. It is only called for enabled buttons (See `btn_text`). Where `btn` is the button id, with a maximum range of `0` to `2`. (The button name mentioned first in `btn_text` corrosponds to `button id = 0`)

Each page should have an own file in your dialogue folder.

Now create a dialogue data in the same folder, it has the following structure:
```
extends Dialogue_Data

class_name ExampleData

# --- /
# -- / class constructor 
func _init():
	entries = [Page1.new(), ...]
```
Register your pages there.

# What should I do in by `btn_action` method?
What you want.

Some useful calls inside the `btn_action` method:
`SingletonPlayer.exit_dialogue()` - exists the dialogue (means closing the window and reallows movement of the player)
`SingletonPlayer.dialogue.select_page(1)` - selects another dialogue page of YOUR current dialogue. The first page has index 0.
`SingletonPlayer.dialogue.finish_dialogue()` - marks the dialogue as FINISHED. It does not close the dialogue window.

# How to link my dialogue to a npc?
This is rather simple. Navigate to dialogue/Dialogue_Registry.gd in the overworld. You find a dictionary there:
`static var npc_dialogues = {1: ExampleData.new()}`
Insert your dialogue data there. The key is the npc id
# How to open the dialogue?

If the dialogue is properly registered with the npc, press e there and it will open, if it is not finished yet. To check if an dialogue is finished, just read the dialogue data from the dictionary and check if `finished` is set to true. (`finished` is a field in the dialogue data super class Dialogue_Data in dialogue in the overworld)
# Dialogue picture
There is no option disabling it. Include a picture of your choice. The size is 150 x 150 px. Including "" leads to the default picture (green.png in the assets dictionary)

# Limitations 
The dialogue system allows (tested on a full hd machine) 144 characters in the content

Disregarding the limitation leads to a broken window.


Arch-enemies/overworld/entities/Player/Player.gd
```diff
@@ -36,7 +36,7 @@ func _physics_process(delta):
 
 func player_movement(delta):
 	# disallow movement when in dialogue
-	if SingletonPlayer.is_in_dialogue:
+	if SingletonPlayer.navigation_in_dialogue():
```
ScatteredDrifter, line 39: reasonable yes :)

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
 
 # --- / 
 # -- / Methods for NPC interaction 
+# register here your dialogue, key is npc id 
+@onready var npc_dialogues: Dictionary = {1: ExampleData.new()}
```
ScatteredDrifter, line 288: I would write it down a little differently to make it easier to set a new value, and maybe make it more readable: 
```
var npc_dialogues: Dictionary = { 
   1: ExampleData.new(), 
    ....
} 
```

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
...
+@onready var dialogue: Dialogue = Dialogue.new()
```
ScatteredDrifter, line 289: this is used to contain all the information about dialogue interaction right? 
So information like :
- whether the player is in a dialogue right now? 
- active dialogue id 
- active npc id 
? 
I took those from the definition in Dialogue.gd

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
...
 
-# takes npc_id, sets is_in_dialogue to true, 
-# queries dialogue to display from given id
-# TODO handle dialogue and query dialogue from given npc_id
+# checks whether a npc has a dialogue linked to it
+func has_dialogue(npc_id:int) -> bool:
```
ScatteredDrifter, line 292: good ^

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
...
+	return npc_dialogues.has(npc_id)
+	
+# takes npc_id, enters the dialogue if one is linked to the current npc_id
 func enter_dialogue(npc_id:int):
+	if not has_dialogue(npc_id):
+		print("WARNING: TRIED TO FIND AN NON EXISTING DIALOGUE FOR NPC", npc_id)
+		return
```
ScatteredDrifter, line 299: nice, readable ^^ !

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
...
 	
+	var data : Dialogue_Data = npc_dialogues[npc_id]
+	dialogue.enter_dialogue(data, npc_id)
+
+# returns whether the dialogue has been finished.
+# finished is not just given by closing the dialogue window.
```
ScatteredDrifter, line 305: good explanation, maybe also add it to the wiki once it has been approved!

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -315,20 +284,37 @@ func save_profile_configuration():
...
+		return false
+		
+	return npc_dialogues[npc_id].finished
 
-	is_in_dialogue = true 
-	active_dialogue = npc_id
-
-	pass
+# returns true when the player is currently in dialogue
+func navigation_in_dialogue() -> bool:
```
ScatteredDrifter, line 315: makes sense, good interface for the dialogue structure ^

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -186,37 +186,6 @@ func add_npc_instance(npc_id:int,npc_object:NPC_interaction):
 
 func obtain_npc_object(npc_id) -> NPC_interaction:
 	return dictionary_npc[npc_id]
-
```
ScatteredDrifter, line 189: I agree that those can be removed, they were primarily used to add a dummy anyway, thx :)

Arch-enemies/overworld/scripts/Npc.gd
```diff
@@ -167,7 +167,7 @@ func check_quest_condition() -> bool:
 					return true 
 				return false 
 			NPC_interaction.Quest.NPC:
-				if SingletonPlayer.check_npc_state(required_npc_id): 
+				if SingletonPlayer.check_dialogue_finished(npc_id):
```
ScatteredDrifter, line 170: reasonable, nice :)

ScatteredDrifter (commented):
I went through your implementation and annotated some aspects to consider. 
Its not much and mostly small nitpicks, yet it would be nice if you took a look at it :) 

I will approve this PR once I've locally tested and checked your implementation, I only ran through the code yet

Arch-enemies/singleton_scripts/singleton_player.gd
DerSimi: Whether the player is in dialogue, the active dialogue (so the active npc id), yes

ScatteredDrifter: thx!

Arch-enemies/overworld/dialogue/Dialogue.gd
```diff
@@ -0,0 +1,59 @@
...
+# --- /
+# -- / class constructor 
+func _init():
+	pass
+
+func enter_dialogue(dialogue:Dialogue_Data, npc_id: int):
+	current_npc_id = npc_id
+	is_in_dialogue = true 
+	active_dialogue = dialogue
+	SingletonPlayer.add_quest_string(npc_id)
```
ScatteredDrifter, line 22: this will add the quest to the ui, we may also implement a method to remove it once again. --> however that depends on the specific dialogue I would say

ScatteredDrifter (approved):
Tested it locally and found some minor issues ( like wrong (static) alignment of text-box, setting the quest to ui)  which you fixed afterwards. 

Other than that your implementation satisfies the requirements and is highly modular/customizable, nice! 

You could potentially add some more internal documentation to the dialogue-system but its names and methods are descriptive too.

PR #268, opened 2024-01-14
overworld | improve save \ loading information ( to/from singleton)

### LINKED ISSUE #267: overworld | improve saving / loading from singleton
## Motivation
Whenever we leave the overworld in any variant it will save the current state of the player in the save-file. 
Furthermore whenever entering the overworld the same state will be loaded from the save-file. 
This is undesired as it adds overhead for information that are constantly available ( the singleton!).
More importantly: Because the save-management is not implemented well / completely yet and misses a lot of information loading into a world will create issues, like not displaying the correct inventories ( of the ui ). 

## Implementation
change small parts in the implementation of overworld to accomplish correctly storing and loading from the singleton.

ScatteredDrifter:
## Motivation
resolves #267 
but primarily adds a better flow of data to **load** information when **entering the overworld**. 

**Now**:
- we only save the state to **file** when the game is closed 
- we only load from **file** when the game is started ( and the singleton is instantiated

- everything related to the overworld draws its information from the **singleton** now --> Player, Camera-Zoom, Overworld-Ui ...
- those objects also **update the singleton** whenever a value changed ( like zoom for example )
- added functionality to **save coordinates** when entering a bridge-scene
- added functionality to load coordinates, zoom and inventory from **singleton** when entering the overworld

## What was changed/added
All of the above: 
-> Those changes spread across many files in our overworld directory. 

1. added a method to send a signal to **overworld-ui** helping to load the inventory from the singleton into the overworld-ui --> and then displaying it correctly upon entering a scene
2. moved `saved_state()` from save_management to singleton and renamed it to `save_player_state()`. All its content is now drawn from the singleton itself 
3. added method `save_game()` to allow saving the whole game by calling this from singleton --> could also be done with reference to the savemanager but this provides **an interface** that could do additional tasks too. ( show a tween to close the game or whatsoever ... ) 

## Testing
**You may have to delete your save-file**  --> see the wiki to find out how :)  
Visually not much changed but you should now be able to: 
- enter the pause menu, close it and be in the same position as before ( without saving to the file! ) 
- enter the bridge-game, solve it ( no exit button yet! ) and be in the same position as before
- the zoom should restore automatically too!

### DIFF
--- a/Arch-enemies/overworld/entities/Player/Player.gd
+++ b/Arch-enemies/overworld/entities/Player/Player.gd
@@ -8,7 +8,6 @@ signal saved_player()
 # -- / default values for visualization
 
 @export var SPEED = 100
-@onready var zoomlevel:Vector2 = SingletonPlayer.get_player_zoom()
 @onready var anim:AnimatedSprite2D = $AnimatedSprite2D
 
 # for correct animation: save last direction walked in, and if sprite was flipped
@@ -25,9 +24,11 @@ signal saved_player()
 @onready var interactionLabel = $interactioncomponents/InteractLabel
 
 func _ready():
-	#debug 
-	$Camera2D.player_object = self
-	#update_interactionLabel()
+	# setting camera zoom
+	var camera_reference = $Camera2D
+	camera_reference.player_object = self
+	camera_reference._set_current_zoom(SingletonPlayer.zoomlevel)
+	position = SingletonPlayer.player_coordinate
 
 func _physics_process(delta):
 	player_movement(delta)
@@ -126,6 +127,7 @@ func execute_interaction():
 					set_interactionLabel(interaction_data["text"])
 					
 				var bridge_edge:SingletonPlayer.BridgeEdge = interaction_data["bridge_edge"]
+				save_player()
 				enter_bridge_scene(bridge_edge)
 				
 			Interactable.InteractionType.ITEM: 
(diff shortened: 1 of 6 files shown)


DerSimi (changes requested):
All in all, this should be fine.

A few things to consider:
- What's the need of the saved_player Signal in Player.gd?
- The game is saved if you exit it via the exit button in the pause menu, but not when closing the window by pressing X on windows
- maybe we could also store if a dialogue was already finished
=> Store an additional dictionary with key npc id and value bool, insert it into the dialogue when starting the game

ScatteredDrifter:
> maybe we could also store if a dialogue was already finished 

this is not done here. 
Like we dont save many information right now ( animal inventory, state of world and more ) 

This PR does **not aim to update save-management** but rather resolve **problems with loading / exiting scenes** and loading data from the singleton.

ScatteredDrifter:
> One check failed, why? 

automatically adding it to some kanban failed.

ScatteredDrifter:
> The game is saved if you exit it via the exit button in the pause menu, but not when closing the window by pressing X on windows. See https://docs.godotengine.org/en/3.2/tutorials/misc/handling_quit_requests.html?highlight=quit
@DerSimi 
Fixed this in  53a3e

ScatteredDrifter:
> What's the need of the saved_player Signal in Player.gd?

Its been a remnant of a previous implementation. 
I removed it now.

PR #269, opened 2024-01-15
overworld | add Npc to list all quests 

### LINKED ISSUE #256: overworld | "Deer-NPC" that contains and tracks the current quests
## Motivation
Proposed motivation for this: 
> This animal will act as **tracker for active quests** so that we dont have to display them in the ui constantly 

Think about this proposal --> it might introduce issues with people understanding / forgetting about their quests, so constantly displaying them would be better in terms of **gameplay** 

However this **NPC** should also:
> only be talked to ( or its quest to be resolved ) upon solving **all other quests** in our game. 

## Implementation
To achieve this: 
- We ought to collect all  quests from "available npcs" -> the ones loaded in our singleton upon loading the overworld
- set a new quest for the npc denoted --> add new option to enum
- update checking conditions for this new quest-type

ScatteredDrifter:
~~// This PR is in progress, the current state does not denote its final state // ~~

## Motivation

resolves #256 

Allows **one single NPC** to have all quests of NPCS available as its own quest --> its only solved if we have solved al lthe others too.

## what was changed/added :

This pr adds a new Quest-Type --> MANY, and its primarily / only  used to allow one NPC to contain all quests available. 
We could technically rewrite it a little more to allow many NPCs to simply have more **quests at a time**, however this case is not considered in the current implementation of the game we strive for ( see wiki for details of our final implementation ). 

I also added the option to **allow** having a bridge as reward --> so that we could **for example** get to a new island once this Npc ( with all quests ) is satisfied in their conditions. 

I had to add a **CONSTANT** that denotes the NPC which will hold all quests as its quest --> its stored in the singleton `QUEST_TRACK_NPC_ID`. 
Whenever we interact with them they are going through the list of Npcs and queries their string-representation of  a quest and maps the state to it. 
We then list those quests in the dialogue box 

This "datastructure" also allows us to check whether the conditions have been met ( all quests without the this one have been solved ) and thus check whether to return the right reward or not.

Whenever we are receiving a valid **bridge-reward** it will not be added to the adjacency-list and update the bridge to be built upon reloading the scene. --> **This should be improved I guess**, or maybe not if it does not make sense regarding story ( why should it suddenly be there? ) 

**NPC-Dialogue**:
I've also added a new NPC-Dialogue, which is not done yet, but contains the quests unsolved in the first page. 
--> **Issue**: This list can get large and thus requires a larger dialogue-box. @DerSimi could you implement this maybe? 

## Testing
So far it works fine. 
There are smaller nits and all, which have to be taken care of. ( I will clean up and refine code once I have more time ) 

You can try talking to the Deer in the upper left corner. It should ( ifthe dialogue-box has been fixed by @DerSimi  ) contain all quests currently undone. 

You can also test talking to the snake -->
Only after you've finished the quest **and** finished the dialogue the corresponding quest will be deleted from the Ui!


mango-gremlin:
I'm not sure whether this pr is finished or not, so disregard this if this will still be implemented.

I tried talking to the deer in the corner, it listed a quest - nice. I tried to talk to any other animal but no dialogue popped up? Only the snake had a blank dialogue box. After talking to the spider, then to esther, then to spider again, the quest doesnt disappear in the deer dialogue either, unless i am doing something wrong. So currently this does not work as described.

ScatteredDrifter:
@mango-gremlin 

> I tried talking to the deer in the corner, it listed a quest - nice. I tried to talk to any other animal but no dialogue popped up? Only the snake had a blank dialogue box. After talking to the spider, then to esther, then to spider again, the quest doesnt disappear in the deer dialogue either, unless i am doing something wrong. So currently this does not work as described.

This is because the dialogues have not been implemented yet. 
One ought to be able to speak to them and finish their Dialogue --> only then it will query a potential quest ( and list it ) and further only when a dialogue has been finished we register that state to be "talked to person x" 

So yea its not working because the dialogues are not implemented yet. 

--> the **Snake** should not have a "blank" dialogue, theres sample text. 
Furthermore its quests gets added and only when one has finished the dialogue it will then get removed again ( because its condition was met --> dialogue done **and** quest done

ScatteredDrifter:
>Furthermore: the bridge reward is nice, although not in the story. What the deer npc needs to have though, is that if all its tracked quests are completed, you get 1-2 or any number of deers - the deer joins you if you help everyone else. That should also be implemented.

I see, was not sure whether it makes sense to have this capability ( like allowing player to traverse to final island if all quests done or such ).
Will add the capability to add more reward animals instead of one

ScatteredDrifter:
@mango-gremlin 

>Furthermore: the bridge reward is nice, although not in the story. What the deer npc needs to have though, is that if all its tracked quests are completed, you get 1-2 or any number of deers - the deer joins you if you help everyone else. That should also be implemented.

Was added now, should work as required.

## Testing 
You can test whether it works by: 

1. Testing a **solvable quest**: set required bridge start to  0 and required bridge dest to 1 for the **snake** - 
This state should be the **default one** at least for this test-scene (( once it has been implemented it will likely be different anyway ) 
Now you can: 
- set the rewardType to Animal 
- set the rewardType to Item
--> For both you should received the **set reward_amount** after talking to the NPC! 

2. Testing an **unsolved quest**: setting required bridge start to  1 and required bridge dest to 4 for the **snake** - Rubert 
Now you can: 
- set the rewardType to Animal 
- set the rewardType to Item
--> For both you will not receive anything after talking to the snake

mango-gremlin:
Thank you!
Now while trying to test it, the Game crashed and gave me this error:
image

is this something related to the save files, do I have to delete the old one again?

ScatteredDrifter:
@mango-gremlin Fixed the issue. 
It occurred because I did not consider updating the case of receiving a bridge as reward ( --> which is only done in the deer right now ).
Should be handled correctly now.

mango-gremlin:
Alright I tested it again, thanks for the quick fix.

I have a couple of things to note, maybe its just me not understanding but:
- i followed your steps for testing, that worked, but
- i tried to get the deer to track the quests: i set the snake to want honey, and the item next to the snake to be honey. I talked to the snake -> quest that it wants honey was displayed in the UI -> but not when talking to the deer; since I have to implement the overworld: is there another step I missed, what did I do wrong here?
- also, it just occured to me: i think it would be better if the deer displayed all quests even before you pick them up, or have some other option to say "you did all the quests you accepted, but there are still quests you need to do" (without listing ones you havent accepted yet, so you dont have to re-implement it) or something.
- also, when i gave the snake the honey, the dialogue did not appear again, I just received the items without any indication. To solve this, I could in theory make another dialogue, that only displays when you have the required item in your inventory, right? How would I go about this?
- also: i think the point you mentioned about also tracking the quests in the ui somewhere is a good idea, but I would still also use the deer npc as well.

ScatteredDrifter:
> also: i think the point you mentioned about also tracking the quests in the ui somewhere is a good idea, but I would still also use the deer npc as well.

I do agree, yes.

ScatteredDrifter:
> also, it just occured to me: i think it would be better if the deer displayed all quests even before you pick them up, or have some other option to say "you did all the quests you accepted, but there are still quests you need to do" (without listing ones you havent accepted yet, so you dont have to re-implement it) or something.

It will, the initial problem was, that the dialogue-box was not large enough and thus it was not possible to display all of them. 
**Simon** pushed his template to dynamically generate an amount of pages and I will incorporate this later ( likely this day ! ) 
Until then the quests would all be listed ( yet not completely visible ) but it should technically work.

ScatteredDrifter:
I've changed the behavior in the last three commits: 

1. The deer will **only track quests** that we know off --> requires us to speak with an animal 
2. Whenever we talk to it its state will update and it will check whether we can solve a quest --> it will add " solvable !" to the quest-string --> **this could be improved with an icon or such** --> we can just use the received bool to manipulate visualization or such 
3. Only if **all(meaning within the overworld)** quests were done it will issue itself as "complete" --> we could change that but that was the requirement afaik. 

To allow testing that: 
-> added dummy dialogues to the other npcs 
-> improved naming because I confused dialogue.enter_dialogue() with SingletonPlayer.enter_dialogue() --> its no prepare_dialogue()
added some methods or modified them to accommodate the changed values and interaction.

mango-gremlin (approved):
I tested it now and it works, thank you very much!

PR #271, opened 2024-01-16
bridges | added pause menu to new bridge level

### LINKED ISSUE #174: bridges | embed pause menu from overworld into bridge level

aylinfi:
## Motivation
closes #174 

The pause_menu wasn't accessible in the new bridge level, so it needed to be added.

## What was changed/added
The pause_menu isn't accessed by changing the scene like before, but by having a pause_menu node inside of the scene.
Opening it by pressing "esc" just changes the visibility of the node, so it can be opened and closed.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/7ab09900-117f-4232-b3f7-9ba650c068e3

You can open and close the pause_menu by pressing "esc". 
Interestingly enough it looks different when it actually changes the scene like in the overworld eventhough it's the same menu.

## Additional Information
You can still move the animals when the pause_menu is pressed. This is also the case in the goal_menu, so if we want to change this, both should be edited.

Both of the menus should also be updated in ther visual apperance by the UI team. Trying to center the menus is kind of frustrating since the bridge scene doesn't seem to actually fill the screen outline of godot, so the automatic centering options of godot also don't work.

### DIFF
--- a/Arch-enemies/bridges/scenes/Grid.tscn
+++ b/Arch-enemies/bridges/scenes/Grid.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=34 format=3 uid="uid://bl1jporj7hosu"]
+[gd_scene load_steps=35 format=3 uid="uid://cjsjvpeuuqot"]
 
 [ext_resource type="Texture2D" uid="uid://d0o2q6fijmd1" path="res://assets/art/tilesets/bridge_tileset/bridge_background_color.png" id="1_uowbf"]
 [ext_resource type="Script" path="res://bridges/script/Camera.gd" id="1_vkapa"]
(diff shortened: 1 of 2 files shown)


Kabooni (approved):
It works, thank you!
I personally would prefer if the player couldn't move while some kind of menu is active. If you want, you can implement it because it's a small change, but if you don't want to, the UI team can do it later while finishing the menus appearance-wise.

Raining-Cloud (commented):
https://github.com/mango-gremlin/arch-enemies/assets/59502349/84d64426-1374-4e50-a380-d14ef18babc3

Kabooni:
@Raining-Cloud I'm not quite sure what you mean. The fox can't jump over the deer, that's not a bug, or is it something else?

Kabooni:
> Well, shouldnt a pause menu actually pause the game ? Right now its just a ui component being displayed 😅
Ah yes, I agree with you. I've already written a comment on that topic, but I can understand why @aylinfi did it, as the goal menu has the same functionality.

mango-gremlin:
I would say pausing the game (eg fox not moving etc) is not a priority but if @Kabooni  has time she can implement it when the pause menu ui is being implemented.
So should we merge this then?

PR #273, opened 2024-01-16
overworld | bridge remove empty inventory to overwrite state saved in singleton

ScatteredDrifter:
## Motivation
merging the bridge inventory and also updating the save management --> while also setting it with default values 
the **debugging function to set an arbitrary animal-inventory** was overwritten by the save management that loads a state upon instantiation. 

This is now fixed, by removing the empty inventory from the save management. 
Furher I added a flag that indicates where to set the inventory once it can be saved / loaded accordingly.

## Testing
Tested by @aylinfi  live


aylinfi (approved):
I did test it and it worked now 👍

PR #275, opened 2024-01-16
bridges | check inventory before dragging

### LINKED ISSUE #264: bridges | clicking on an animal that is not placable keeps the helping-grid visible
## Description

**With the new inventory system**  it is possible to click on an animal without it being available ( amount in inventory is 0 ). 
Doing that will **display the helping-grid** yet not remove it after it has been dropped ( or simply not clicking it anymore because you cannot place an instance of the animal either ). 

**To Reproduce**
1. start Grid-scene
2. click on Spider ( in current branch its always empty ) ( this would also happen with other animals that are empty ) 
3. the grid appears but does not get removed 

**Expected behavior**
it should not be visible at all or disappear shortly after clicking the animal

**Screenshots**
This denotes the issue ( its not aligned because the size of this window changed ) 
image

## Solution Proposal

I can imagine that the **method for dragging** an animal will automatically set the **grid to be visible** yet only disables it after the **animal has been placed or dropped**.

**Now** with the addition of an inventory its possible for an animal not to have any amount left. 
There it will display the grid but the **breaking condition** will not disable it again. 

> This should be easy to fix, however I'm not familiar with the implementation enough to fix it easily

Kabooni:
## Motivation
closes #264

Because we start dragging an animal before checking the inventory, the helping grid appears even though you might not be able to drag the animal.

## What was changed/added
Move the inventory check at the top of get_drag_data.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/57258671/252e4ddd-e2e3-4a5a-9fa4-e6f95ded00ac

Works with the test inventory.

### DIFF
--- a/Arch-enemies/bridges/script/grid_drag.gd
+++ b/Arch-enemies/bridges/script/grid_drag.gd
@@ -31,32 +31,32 @@ func _process(delta):
 
 func _get_drag_data(at_position):
 	if Global.drag_mode:
-		#When we try to drag something we first need to know what the grid looks like
+		# First we need to check if we have some animals of the requested type left in the inventory
+		var animal = tooltip_text 
+		# Due to we use the tooltip we need to check if the string is valid... this is not secure for typos!
+		var animal_type:Animal.AnimalType = Animal.string_to_type(animal)
+		if animal != "" and Grid_node_reference.start_animals[animal_type] <= 0:
+			return
+		
+		# When we try to drag something we first need to know what the grid looks like
 		need_grid.emit()
 		#Then we tell the grid that we are currently dragging something
 		#This allows us to see FORBIDDEN and ALLOWED zones
 		is_dragging.emit()
 		
-		#We create the data we want to transmit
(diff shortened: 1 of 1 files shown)


marcel-iv (approved):
Code looks good, and everything seems to work as intended.

The empty return statement in line 39 looks a bit weird, but other options don't seem to work from what I've tested briefly, so its fine

Kabooni:
> sorry but do you know why you can't grab the spider?

... because its inventory count is 0?

Kabooni:
> Code looks good, and everything seems to work as intended.
> 
> The empty return statement in line 39 looks a bit weird, but other options don't seem to work from what I've tested briefly, so its fine

As far as I know, you need it to prematurely end the function. Like _break_ in loops.

mango-gremlin:
> > sorry but do you know why you can't grab the spider?
> 
> ... because its inventory count is 0? I just tried to prove that the helping grid won't appear while trying.

ah right, excuse me, i missed that

PR #276, opened 2024-01-16
Extended squirrel logic for shallow

### LINKED ISSUE #263: bridges | placing squirrel after deer has been placed in shallow water
## Description

After placing a deer into shallow water its possible to place an animal to the remaining space ( i.e. a squirrel as of right now ) 
It was described in #231  already.

Pictured below 
image

**To Reproduce**
1. start Grid-scene
2. place a deer into shallow water leaving a free space left /right of it 
3. place a squirrel into water 

**Expected behavior**
This should not be possible because we limit animals to be placed within this area. 

## Solution Proposal
I dont know really i'm not familiar with the implementation much, sorry.

marcel-iv:
## Motivation
closes #263

## What was changed/added
Updated the logic for squirrel, so it accounts for SHALLOW, which means setting the `is_free` to false when any part of the squirrel would end up in shallows. 

## Testing
As you can see, the illegal placement as described in #263 is no longer possible.

https://github.com/mango-gremlin/arch-enemies/assets/104799849/3f726aed-3f50-44e1-9228-381725aa9a5e

As you can see, the squirrel still should work elsewhere

https://github.com/mango-gremlin/arch-enemies/assets/104799849/416c3cc2-439d-4c9f-a75a-8bebd9e68dd0

### DIFF
--- a/Arch-enemies/bridges/script/grid_drag.gd
+++ b/Arch-enemies/bridges/script/grid_drag.gd
@@ -184,10 +184,10 @@ func is_squirrel_allowed(pos):
 			is_allowed = true
 		#Is there space for the animal?
 		if(pos_Type == ENTITY_TYPES.FORBIDDEN 
-		or pos_Type == ENTITY_TYPES.GROUND 
-		or pos_Type == ENTITY_TYPES.WATER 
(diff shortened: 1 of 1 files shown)


Kabooni (approved):
lgtm, thanks!

PR #278, opened 2024-01-16
bridges | disable all actions when menus are open

### LINKED ISSUE #272: bridges | opening menus should prohibit player from moving
## Motivation
As discussed in #271 the player can still move around even while the pause menu is open.

## Implementation
Create a new global variable that tracks if a menu is opened and limit player movement based on it.

aylinfi:
## Motivation
closes #272 

All actions where still fully functional when the pause or goal menu was open. They should be paused/disabled instead.

## What was changed/added
Added global variables to keep track of when menus are open and changed functions to be dependent on these variables.
Also wrote new global function to change visibility of all ui elements which also disables them when invisible.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/8b8cde02-88a7-4ce1-a6b8-cae678a56fad

## Additional Information
Remind me to add similar functionalities to overworld menu.


Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -125,9 +125,9 @@ func add_to_animal_inventory(new_animal:Animal.AnimalType, quantity:int = 1):
 # used for debugging only
 func set_test_animal_inventory() -> Dictionary:
 	var inventory:Dictionary = Animal.init_animal_inventory()
-	inventory[Animal.AnimalType.DEER] = 1
-	inventory[Animal.AnimalType.SNAKE] = 1
-	inventory[Animal.AnimalType.SQUIRREL] = 1
+	inventory[Animal.AnimalType.DEER] = 5
```
ScatteredDrifter, line 128: this shouldn't be changed ( normally )

mango-gremlin (changes requested):
this looks great, but maybe we should refrain from further using the global script, as it is kind of bad practice to use this kind of thing? Instead maybe we could save it in the singleton, in which case the same stuff can also be used for disabling movement when in the menu in the overworld. Btw, is this something that you implemented here as well? Because it should also work in the overworld in the same way with disabling movement.

Arch-enemies/overworld/ui/menu/menu/pause_menu.gd
```diff
@@ -13,11 +13,19 @@ func _process(delta):
 
 
 func _on_overworld_pressed():
+	# reset global variables
+	Global.drag_mode = true
```
ScatteredDrifter, line 17: wouldn't it be beneficial if we extracted that to some method? 
something like:
```
reset_global_bridge_state(state:bool)
	Global.drag_mode = true
	Global.menu_mode = state
	Global.goal_reached = state
```

ScatteredDrifter: ( would reduce the amount of ducplicated code and make it easier to read )

Arch-enemies/bridges/script/toggle_mode_button.gd
```diff
@@ -1,10 +1,11 @@
 extends CheckButton
 
+# FIXME "res://bridges/scenes/toggle_build_fox_mode.gd" replaces this skript
```
ScatteredDrifter, line 3: maybe remove it ( or is there to provide information afterwards? )

Arch-enemies/bridges/script/player.gd
```diff
@@ -18,46 +18,49 @@ func _ready():
...
-	direction = Input.get_vector("left", "right", "ui_up", "ui_down")
-	if direction:
-		velocity.x = direction.x * speed
-	else:
-		velocity.x = move_toward(velocity.x, 0, speed)
-	
-	move_and_slide()
-	update_animation()
-	update_facing_direction()
+	if not Global.menu_mode:
```
ScatteredDrifter, line 21: we could use less indentation by inverting this condition 
```
 if Global.menu_mode:
     return
```
might make it easier to read :)

Arch-enemies/bridges/script/goal_menu.gd
```diff
@@ -1,13 +1,22 @@
 extends Control
 
 func _on_retry_button_pressed():
-	get_parent().get_node("Player").reset_player()
+	# reset everything
+	var grid = get_parent()
+	grid.get_node("Player").reset_player()
+	Global.change_ui_visibility(true, grid)
 	Global.drag_mode = true
+	Global.menu_mode = false
```
ScatteredDrifter, line 9: would once again advise extracting this in a separate method as denoted below

ScatteredDrifter (commented):
ran through your changes and made some minor annotations. Maybe you could answer / go over them. 

Other than that your changes seem reasonable :)

Arch-enemies/singleton_scripts/singleton_player.gd
ScatteredDrifter: considering that its for testing purposes only and that we require some animals to test a level, thats fine  //

aylinfi:
There is an error when reaching the goal and pressing continue, but this is related to bridge_connections/edges which I am not at all familiar with.

Kabooni:
> There is an error when reaching the goal and pressing continue, but this is related to bridge_connections/edges which I am not at all familiar with.

Did you test it while directly starting in grid.tscn? As the levels are now connected to the main overworld, you need to start there and enter the level for the bridge_edge to be set correctly. Without doing that, the level doesn't know its own bridge edge.

Arch-enemies/overworld/ui/menu/menu/pause_menu.gd
```diff
@@ -14,18 +14,14 @@ func _process(delta):
...
 	# TODO improve  by constantly saving the previous scene in a variable!
 	get_tree().change_scene_to_file("res://overworld/main_scene_overworld.tscn")
 
 
 func _on_exit_pressed():
 	# reset global variables
-	Global.drag_mode = true
-	Global.menu_mode = false
-	Global.goal_reached = false
+	get_parent().reset_modes()
```
ScatteredDrifter, line 24: this does not work globally! 

If I were to call execute the menu in the overworld the game crashes because its parent ( which is not grid! ) does not contain this method.

Arch-enemies/bridges/script/player.gd
```diff
@@ -21,37 +21,44 @@ func _ready():
...
-
-		# Get the input direction and handle the movement/deceleration.
-		# As good practice, you should replace UI actions with custom gameplay actions.
-		direction = Input.get_vector("left", "right", "ui_up", "ui_down")
-		if direction:
-			velocity.x = direction.x * speed
-		else:
-			velocity.x = move_toward(velocity.x, 0, speed)
-		
+	if get_parent().menu_mode:
```
ScatteredDrifter, line 24: maybe you can query the parent once and then reference it every time. looks better regarding code quality :)

Arch-enemies/bridges/script/grid.gd
```diff
@@ -416,19 +420,36 @@ func tile_update(pos, current, next):
...
 # Called every frame. 'delta' is the elapsed time since the previous frame.
 func _process(delta):
 	if(Global.currently_dragging and Input.is_action_just_released("click")):
 		make_invisible()
 		
 	# pressing "esc" opens the pause-menu
-	if Input.is_action_just_pressed("open_menu") and not Global.goal_reached:
-		var pause_menu = get_parent().find_child("pause_menu")
+	if Input.is_action_just_pressed("open_menu") and not goal_reached:
+		var pause_menu = get_parent().find_child("bridges_pause_menu")
```
ScatteredDrifter, line 445: I would once again set an \@export link to help setting this value instead of querying it relative from the node-structure. In case the hierarchy changes this would break again

ScatteredDrifter (changes requested):
thx for updating this repo. 
The menu does not work in the overworld as it tries to query the method "reset_modes()" from a parent, however this structure is not present when executing in the overworld.

Arch-enemies/overworld/ui/menu/menu/pause_menu.gd
aylinfi: i fixed this now

Kabooni (approved):
As everything left is refactoring, I would propose we merge this pr and do the refactoring later with the rest. I would love to have this feature for playtesting.

ScatteredDrifter (approved):
after you've fixed the issues described ( primarily game crashing when opening the menu in the overworld ) it works well, thanks :)

mango-gremlin (approved):
lgtm!

PR #279, opened 2024-01-16
bridges | update singleton after level was solved

### LINKED ISSUE #250: bridge | method to return from bridge-level to overworld
## Motivation
is important to complete the game-loop and be able to leave the level after:
- one completed it 
- did not complete it but wants to leave

## Implementation

We **internally track** whether the game was solved already or not ( could be a simple boolean ). 
upon "leaving the scene" --> i.e **clicking a button** in the _pause menu_ or the _complete menu_ we should do the following:
1. **check** whether level was solved or not 
2. **if it was solved**: --> update bridge-connection, by giving back the correct bridgeEdge ( that denotes this bridge-level ). Use `SingletonPlayer.add_bridge_connection()` to add this new connection! 
3. **if it was solved**: --> update **inventory** of player --> animals were used thus we have to send the internal inventory we have to the singleton again: `Singleton.set_animal_inventory()` to update the active inventory!
4. **if it was not solved**: do none of the above because nothing changed! 
5. **in both ways**: --> change the active scene back to **overworld**

Kabooni:
## Motivation
closes #250

After the level is finished (which happens when entering the _goal area_), we want to update the global inventory and bridge connection list.

## What was changed/added
When clicking the "Continue" button in the goal menu, the SingletonPlayer inventory is updated and the necessary bridge-edges are added.
Because while returning to the overworld we will not change the level, nothing else has to be changed.

## Testing
https://github.com/mango-gremlin/arch-enemies/assets/57258671/d793841a-b36c-4e44-8de9-7f631c33e4b4
Through print statements before and after you can see that the bridge edge was added correctly.
grafik

## Additional Information
For this implementation to work the current level must exist in SingletonPlayer.bridge_level_scenes. If there is a better solution to find out which BridgeEdge we are using in the current level, let me know.

### DIFF
--- a/Arch-enemies/bridges/script/goal_menu.gd
+++ b/Arch-enemies/bridges/script/goal_menu.gd
@@ -12,16 +12,8 @@ func _on_retry_button_pressed():
 func _on_continue_button_pressed():
 	# Update global inventory
 	level_solved.emit()
-	# Add bridge edge to Singleton, I used the values from Singleton.bridge_level_scenes
-	#print("Build bridges for finish: ", SingletonPlayer.bridges_built)
-	SingletonPlayer.add_bridge_connection(get_bridge_edge())
-	#print("Build bridges after finish: ", SingletonPlayer.bridges_built)
+	# Add bridge edge to Singleton
+	var current_edge = SingletonPlayer.get_current_bridge_edge()
+	SingletonPlayer.add_bridge_connection(current_edge)
+	SingletonPlayer.set_current_bridge_edge(null)
 	get_tree().change_scene_to_file("res://overworld/main_scene_overworld.tscn")
-
-func get_bridge_edge():
-	var levels = SingletonPlayer.bridge_level_scenes
-	for level in levels:
-		var path = levels[level]
-		if path == get_tree().current_scene.scene_file_path:
-			return levels.find_key(path)
-		# TODO: create new entry?
--- a/Arch-enemies/bridges/script/goal_menu.gd
+++ b/Arch-enemies/bridges/script/goal_menu.gd
@@ -13,11 +13,10 @@ func _on_continue_button_pressed():
 	# Update global inventory
 	level_solved.emit()
 	# Add bridge edge to Singleton, I used the values from Singleton.bridge_level_scenes
-	print("Build bridges for finish: ", SingletonPlayer.bridges_built)
+	#print("Build bridges for finish: ", SingletonPlayer.bridges_built)
 	SingletonPlayer.add_bridge_connection(get_bridge_edge())
-	print("Build bridges after finish: ", SingletonPlayer.bridges_built)
+	#print("Build bridges after finish: ", SingletonPlayer.bridges_built)
 	get_tree().change_scene_to_file("res://overworld/main_scene_overworld.tscn")
-	print("lets go to the overworld now")
 
 func get_bridge_edge():
 	var levels = SingletonPlayer.bridge_level_scenes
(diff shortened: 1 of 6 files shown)


Arch-enemies/bridges/script/goal_menu.gd
```diff
@@ -8,6 +10,18 @@ func _on_retry_button_pressed():
 
 
 func _on_continue_button_pressed():
-	# TODO return to overworld scene
+	# Update global inventory
+	level_solved.emit()
+	# Add bridge edge to Singleton, I used the values from Singleton.bridge_level_scenes
+	#print("Build bridges for finish: ", SingletonPlayer.bridges_built)
+	SingletonPlayer.add_bridge_connection(get_bridge_edge())
+	#print("Build bridges after finish: ", SingletonPlayer.bridges_built)
```
ScatteredDrifter, line 18: besides updating the bridge-connection we should also update the animal inventory --> as we've used animals locally yet have not propagated that change into our singleton ! 
Other than that your implementation seems fine so far.

Arch-enemies/bridges/script/goal_menu.gd
```diff
@@ -8,6 +10,18 @@ func _on_retry_button_pressed():
...
 	get_tree().change_scene_to_file("res://overworld/main_scene_overworld.tscn")
-	print("lets go to the overworld now")
+
+func get_bridge_edge():
```
ScatteredDrifter, line 21: I'm really not sure whether that's a good way to track which level we are in. 

What I was thinking about: 
maybe we can add the current BridgeEdge we are "playing at" to a temporary value upon entering the world ( comes from singleton-side here ). 
I will think about it once more, but maybe you can find  a solution for it too. 

I would argue that querying for the path is somewhat possible but seems risky

ScatteredDrifter (changes requested):
Ran through your changes, thanks for your time / contribution :) 

I have two things to add / talk about:
1. we have to update the player inventory ( for animals ) after leaving the bridge-level
2. while it seems to be working to query for the path of the current scene, I think that we should maaybe temporarily store / show which "bridgeEdge we play at" allowing us to return this as reward instead.

Arch-enemies/bridges/script/goal_menu.gd
Kabooni: It is getting propagated - I used the function set_global_animal_inventory you wrote in grid.gd. That's what the level_solved.emit() is for, to notify the grid.

Kabooni:
I've changed to edge finding approach. There's now a new variable in SingletonPlayer, `current_edge` which is getting set in the player script while entering a bridge level, and retrieved by the goal menu script while setting the solved bridge edge.

I'm not sure if using the singleton for that is the best way, but I'm not really familiar with the overworld code to find a better solution.

Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -163,6 +163,17 @@ var islands_reachable:Array[bool]
 	BridgeEdge.new(0,2): "res://bridges/scenes/Grid.tscn",
 }
 
+# used for the (possibly) active bridge level to find out its own bridge egde
+@onready var current_bridge_edge
```
ScatteredDrifter: I would set this type as well ( will add it in a small commit )

Arch-enemies/bridges/script/goal_menu.gd
```diff
@@ -8,6 +10,10 @@ func _on_retry_button_pressed():
 
 
 func _on_continue_button_pressed():
-	# TODO return to overworld scene
+	# Update global inventory
+	level_solved.emit()
+	# Add bridge edge to Singleton
+	var current_edge = SingletonPlayer.get_current_bridge_edge()
+	SingletonPlayer.add_bridge_connection(current_edge)
+	SingletonPlayer.set_current_bridge_edge(null)
```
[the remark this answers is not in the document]
ScatteredDrifter: changed it

ScatteredDrifter (approved):
after going through your changes once more, I would approve your implementation / changes :)

PR #285, opened 2024-01-18
Bug | leaving overworld updates item inventory instead of animal inventory 

ScatteredDrifter:
## Motivation 
I observed a bug that updates the wrong inventory upon entering the overworld after a level was solved. 
I have not seen it when reviewing the PR, my bad / sorry. 

Additionally I gave some _more reasonable_ values for the test inventory --> so that ones does not have to change it that often. This will likely collide with #278 

## What was changed/added
using the right method in "grid.gd" to update the animal_inventory instead of item_inventory.

## Testing
Well, solve the level and you should observe that the updated inventory should be displayed in the "ui".


mango-gremlin (approved):
tested it by finishing a level and then going back to overworld -> it works! thanks!

Kabooni (approved):
Thanks a lot for noticing it!

PR #287, opened 2024-01-18
Issue280/sound

### LINKED ISSUE #280: Add sound effects
## Motivation
As per request, we should add audio to our game to improve the feel of our game. We specified which sounds we wanted in the High Court meeting, though we can expand the scope if we have suffient time, or upon request in the playtests.

## General Sound Effects - used in both overworld and puzzles
- Start screen music (subtle/ambient)
- Sounds for each animal (used for placement & dialogue)
- Walking 
- Victory sound (quest/level solved)

## Overworld Sounds
- Quest accepted

## Bridge Sound Effects
- Death sound
- Jump sound
- take item from inventory (haptic sound)

## Implementation
As we have a lot of shared sounds in both part of our game, we should think of a way that solves this problem with minimal effort in both implementation and future changes.

Playing sounds in Godot requires a AudioStreamPlayer or AudioStreamPlayer2D (the latter is positional, maybe useful for NPCs), and an attached audio file to play something from code.

marcel-iv:
## Motivation
closes #280 

## What was changed/added
Sound was added as requested in the High Court Meeting, summed up in #280. All the sounds, and the implementation in bridges has been done by @koennixer, credits to him.

## Testing
You can check everything for yourself, and we recommend you do so to check everything for yourself, especially because volume might be drastically different. 

https://github.com/mango-gremlin/arch-enemies/assets/104799849/86434726-c035-4072-8426-e54ff2d5b9b4

https://github.com/mango-gremlin/arch-enemies/assets/104799849/3a1438ad-4e7a-4506-aefd-8c749b8fa815

## Additional Information
!Important! Right now this isn't finished, overworld is missing dialogue sounds, which I will sort out tomorrow probably. Do **not** merge this before everything is included, and before everything is discussed. 

## Bugs found
- Walking sounds depend on the global variable `walking`, which we should refactor anyways, but works for now. However, this causes the walking sound to repeat endlessly in some cases, even though no movement occurs, such as when initating dialogue

## Discussion
Please review all the sounds, as well as the volume, as everything is highly subjective. 

@koennixer mentioned that sound options might be something we could implement in the future, as it shouldn't be too hard (but also associated with new assets, testing etc.)


Arch-enemies/singleton_scripts/singleton_player.gd
```diff
@@ -10,9 +10,9 @@ extends Node
 # Called when the node enters the scene tree for the first time.
 func _ready():
 	pass
-	Savemanager.load_config()
+	#Savemanager.load_config()
 	# TODO add input to set 
-	Savemanager.select_profile("default")
+	#Savemanager.select_profile("default")
```
Raining-Cloud, line 15: why is this commented ?

Arch-enemies/overworld/overworld_sound.gd
```diff
@@ -0,0 +1,43 @@
...
+		"STEP":
+			stream = step
+			playing = true
+		"QUEST_ACCEPTED":
+			stream = quest_accepted
+			playing = true
+		"QUEST DONE":
+			stream = quest_done
+			playing = true
+
```
Raining-Cloud, line 34: same as above

Arch-enemies/bridges/script/grid.gd
```diff
@@ -28,7 +28,7 @@ var save_states = 10
 var state = 0
 
 #for the inventory
-@onready var start_animals : Dictionary = SingletonPlayer.get_animal_inventory().duplicate(true)
+@onready var start_animals : Dictionary = SingletonPlayer.set_test_animal_inventory().duplicate(true)
```
Raining-Cloud, line 31: I guess this was for testing ?

Arch-enemies/bridges/script/bridge_sound.gd
```diff
@@ -0,0 +1,65 @@
...
+	play_sound(sound)
+
+func _on_squirrel_item_play_sound(sound):
+	play_sound(sound)
+
+func _on_player_play_sound(sound):
+	play_sound(sound)
+
+func _on_grid_play_sound(sound):
+	play_sound(sound)
```
Raining-Cloud, line 65: You can directly link the play_sound function => less code

Arch-enemies/bridges/script/bridge_sound.gd
```diff
@@ -0,0 +1,65 @@
...
+			playing = true
+		"SPIDER":
+			stream = spider
+			playing = true
+		"STEP":
+			stream = step
+			playing = true
+		"SQUIRREL":
+			stream = squirrel
+			playing = true
```
Raining-Cloud, line 36: well we don't need to start this topic again, but maybe, just maybe try to avoid such code in future...

Arch-enemies/bridges/script/walking_sound.gd
```diff
@@ -0,0 +1,16 @@
...
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(delta):
+	if $"../Walking_Timer".is_stopped():
+		if Global.walking:
+			playing = true
+			$"../Walking_Timer".start(0.18)
+		else:
+			playing = false
+			$"../Walking_Timer".start(0.18)
```
Raining-Cloud, line 16: lacking of description what this code is for, in general not that many comments. Makes it hard to follow your thoughts

Raining-Cloud (changes requested):
Sound works! 
Wouldn't say its my taste but thats not the topic, I added some comments regarding some issues, probably changes for debugging.

I also commented for Topics like "Clean-Code" this is not for changes but this is general Feedback. Makes life way easier for your team partners :)

Arch-enemies/bridges/script/grid.gd
ScatteredDrifter: oh yea please revert this change :) 
Considering that you can locally set it in the singleton itself, we only have a single point of change

Arch-enemies/overworld/overworld_sound.gd
ScatteredDrifter: I would note that one should use type annotation. 
GdScript does not allow for good documentation ( like preview of javadocs or similiar ) yet it shows the arguments and their types, which helps us here! 

Considering that I might not know what "sound" actually means, we could annotate it to be a string, making it more understandable :) 

Further I would argue that calling it something like `sound_type` or such might make it easier to understand (?)

Arch-enemies/bridges/script/grid.gd
marcel-iv: This was part of @koennixer code, but I'm fairly certain this is just for testing purposes. We'll just revert this when everything else is done

Arch-enemies/bridges/script/bridge_sound.gd
ScatteredDrifter: Agreed, yet we also need to be kind in our code reviews ^^ 

But yea, 
We can see that "playing" is always set to true, so no need to add it to each case. 

String matching might work, but primitive types caaan get confusing or introduce issues --> maybe an enum for this ^?

Arch-enemies/bridges/script/walking_sound.gd
ScatteredDrifter: personally I would set the reference to the timer once and then use this reference! 
once again helps to minimize the amount of code to change in case the path changes, for example. 

so `var timer_reference:DunnoWhichNodeTypeThatIs = $../Walking_Timer`

mango-gremlin (commented):
this looks good and works when tested, great work, thanks!
Just for clarification because I dont know how these sound nodes work: for the other bridge levels, we can just copy the nodes over, and they should work?

koennixer:
Alright. Let's go from top to bottom:

"Wouldn't say its my taste but thats not the topic"

This is not workable advice. If you don't like something tell me what you dislike and what it should rather be. If you don't like 8-bit as sound effect what would you rather have? 16-bit? Realistic? Cartoonish? If you don't like Chiptune what would you rather have? Orchestrated? 8-Bit?

"why is this commented ? (change)"

I don't know why that was commented. It is now uncommented.

"well we don't need to start this topic again, but maybe, just maybe try to avoid such code in future..."

This is not useful. You can be direct about your dislike for things as much as you want, but this is not constructive. What do you want instead? How do you want it done? Fundamentally if you don't want a node for each sound you have to load a new stream into the one you are using. That stream has to be preloaded, ideally this should happen in a Sound Node itself, because that is were it belongs not in the Grid or Drag. Given that obviously other branches of the code will have to tell the sound node to switch the stream and then play it. This can be done two ways, either you will have to, in every single node that is supposed to play sound, load that Node on ready, which I refuse to believe is an efficient way to go about it, or you signal the Node. If you signal the Node you will obviously need to inform it what sound it is supposed to be playing. Ideally this would be done with a Global Enum or something else, but given that the data we are transferring in drag is Strings this seems to be a workable option. Given all that you have given me as critique: 

1. call the function directly, which sure is less lines of code, but I'm less sure about it being more efficient. 

2. Avoid using the match, which is completely unworkable as a critique, because you don't state what you'd rather we use. If it's about the String then state that. If you'd rather we use something that isn't a match then elaborate. If you want us to completely refactor the sound into something else than elaborate.

This does not read as a constructive critique but rather as smugness.

"I guess this was for testing ?"

Yes. And as long as the code is not done I'm gonna leave it as that.

"I would note that one should use type annotation."

Marcel changed it.

"Timer"

Not in regards to a specific quote but the timer has also been changed. Marcel has implemented Evelynns suggestion and I made the timer be loaded on ready, avoid loading it with every delta. These changes are all specific to the grid scene. Marcel can deciede to apply them to the overworld.

"for the other bridge levels, we can just copy the nodes over, and they should work?"

Signal might have to be reapplied, but other than that it should be identical.

mango-gremlin:
> "for the other bridge levels, we can just copy the nodes over, and they should work?"
> 
> Signal might have to be reapplied, but other than that it should be identical.

-> how exactly, and where do we need to reapply the signal?
Or rather, since you two are familiar with it (and have my hands full constructing the final overworld, dialogue etc) - maybe when this is approved and merged, just merge main into aylin's and my branch (feature/final_overworld_and_bridges) and then you can add the sounds to the bridge levels? overworld should work seamlessly, as I just edited the main overworld for the final overworld.

marcel-iv:
> Not in regards to a specific quote but the timer has also been changed. Marcel has implemented @ScatteredDrifter suggestion and I made the timer be loaded on ready, avoid loading it with every delta. These changes are all specific to the grid scene. Marcel can deciede to apply them to the overworld.

Just fyi, both overworld and bridge walking functions the same, as the player is same for both scenes, so this is refactored for both now.

Kabooni (approved):
Sounds great! I won't comment on code quality, as @Raining-Cloud did a lot already.
Personally, I would turn the background music volume a bit down, as it's quite dominant in comparison to the other sounds. In general, the sounds are quite loud, but that is something very subjective and of course, I can just turn the volume down on my own device. So if no one else has the same opinion, it's fine.
And I'm missing some very simple sounds for the buttons, especially for the menus, but as this is not necessary and I think was not part of the issue, it can be skipped.
In general, great work guys! 👍

Raining-Cloud:
> 
> "well we don't need to start this topic again, but maybe, just maybe try to avoid such code in future..."
> 
> This is not useful. You can be direct about your dislike for things as much as you want, but this is not constructive. What do you want instead? How do you want it done? Fundamentally if you don't want a node for each sound you have to load a new stream into the one you are using. That stream has to be preloaded, ideally this should happen in a Sound Node itself, because that is were it belongs not in the Grid or Drag. Given that obviously other branches of the code will have to tell the sound node to switch the stream and then play it. This can be done two ways, either you will have to, in every single node that is supposed to play sound, load that Node on ready, which I refuse to believe is an efficient way to go about it, or you signal the Node. If you signal the Node you will obviously need to inform it what sound it is supposed to be playing. Ideally this would be done with a Global Enum or something else, but given that the data we are transferring in drag is Strings this seems to be a workable option. Given all that you have given me as critique:
> 
>     1. call the function directly, which sure is less lines of code, but I'm less sure about it being more efficient.
> 
>     2. Avoid using the match, which is completely unworkable as a critique, because you don't state what you'd rather we use. If it's about the String then state that. If you'd rather we use something that isn't a match then elaborate. If you want us to completely refactor the sound into something else than elaborate.
> 
> 
> This does not read as a constructive critique but rather as smugness.

Ok so right now it seems that we know via the signal connection which sound gets played by the function getting called and by its parameter. If we remove the multiple _on_[...]_play sound methods we need some sort of checking which sound to play, I would suggest a Dictonary but a match is also a solution in this case. 
But why do we even bother if we already know in the _on_[...]_play sound which sound to play ?

In e.g. this code in bridge_sound.gd should output the same result, at least for what I tested:
```go
extends AudioStreamPlayer2D
var squirrel = preload("res://assets/sound/Squirrel.wav")
var death = preload("res://assets/sound/Death.wav")
var drop = preload("res://assets/sound/Drop.wav")
var deer = preload("res://assets/sound/Deer.wav")
var jump = preload("res://assets/sound/Jump.wav")
var spider = preload("res://assets/sound/Spider.wav")
var step = preload("res://assets/sound/Step.wav")
var snake = preload("res://assets/sound/Snake.wav")


func _on_drag_grid_play_sound(sound):
    stream = drop
    playing = true

func _on_deer_item_play_sound(sound):
    stream = deer
    playing = true

func _on_snake_item_play_sound(sound):
    stream = snake
    playing = true

func _on_spider_item_play_sound(sound):
    stream = spider
    playing = true

func _on_squirrel_item_play_sound(sound):
    stream = squirrel
    playing = true

func _on_player_play_sound(sound):
    stream = jump
    playing = true

func _on_grid_play_sound(sound):
    stream = death
    playing = true
```
And sorry for not providing a alternative instantly in this case, this was not thought for any change, just for the feedback on the general code layout.

Raining-Cloud (approved):

mango-gremlin:
as @Raining-Cloud 's suggestion is refactoring -> do @marcel-iv or @koennixer want to do this right now, or can we merge this, so I can see whether there are any problems with the sound in the new bridge levels?

aylinfi (commented):
I really love the sounds of the different animals!
But I agree with Katja that it's overall very loud (esp bg music). And maybe the bg music and the victory music shouldn't overlap, to give the victory more emphasis.

PR #290, opened 2024-01-19
bridges | update menus for the new levels

### LINKED ISSUE #288: bridges | update menus for new levels
## Motivation
While the menus do work, the size is static and the levels have different sizes. So we should adjust to that.

## Implementation
One possible approach would be to increase the size of the menus

Kabooni:
## Motivation
closes #288

Updated menu design to adjust to different level sizes.

## What was changed/added
The pause and goal menu was extended to the sides to adjust for bigger grid sizes. Extensions up and down can be provided via scaling the pause menu when needed.
Everything was centred to origin to make things easier to place.

## Testing
The new goal menu (pause menu is changed the same way):
grafik

https://github.com/mango-gremlin/arch-enemies/assets/57258671/652d0f51-945b-4065-aa9d-f272be902055
It still works for the bigger levels.

## Additional Information
As seen in the video, everything is kinda squished for the frog hazard level because the zoom is not the same for the x and y axis. It should be considered to change that, or find another way.

### DIFF
--- a/Arch-enemies/bridges/scenes/0_TutorialLevel.tscn
+++ b/Arch-enemies/bridges/scenes/0_TutorialLevel.tscn
@@ -902,6 +902,10 @@ position = Vector2(320, 86)
 
 [node name="bridges_pause_menu" parent="Grid" instance=ExtResource("20_1ny00")]
 visible = false
+offset_left = 193.0
+offset_top = 108.0
+offset_right = 193.0
+offset_bottom = 108.0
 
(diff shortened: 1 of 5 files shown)


mango-gremlin (approved):
looks great, thanks!
did not notice the squish until i saw the level on windows; it didnt squish on linux for some reason. Will fix the squish after i had lunch in the feature/final_overworld_and_bridge branch

PR #296, opened 2024-01-20
fixes stray lowercase sound file

marcel-iv:
should work on any OS now

renamed by accident by me yesterday somehow I think


ScatteredDrifter:
updated to work for linux  ( renamed file ) 
does it work like that for you @marcel-iv @aylinfi ?
if, we can merge /

marcel-iv:
> updated to work for linux ( renamed file ) does it work like that for you @marcel-iv

Considering it also worked on wrong file names for me, yes, it indeed works just as before

Kabooni (approved):
As it seems to work on windows and linux, I will merge this branch

PR #298, opened 2024-01-21
overworld | hide item when it was taking | add states to save progress after scene change

### LINKED ISSUE #179: overworld | item interactions require ID for state management
## Motivation
Item interaction spots require an ID,  to be identified **upon** loading the world state or to update it. 
This is necessary because every Item is only obtainable once and thus requires to be identified --> to set its status or similar 

## Implementation

Although not entirely sure where to position and load it, we need to make this state available in a global scale, because it is required upon loading the game itself. 

1. **each item** should contain a new method to query its state **upon initialization** ( so when ready() is called ). 
2. updating the state of an item_interaction is necessary too --> **updating world state for given item_id**

ScatteredDrifter:
## Motivation
resolves #179 
- also uses this addition to add _state management_ to maintain item amounts after scene changes occurred
- resolves **visibility** of item once its resource was depleted. 
- removes debug message when interacting with an item
-  removes unused variables from item_interaction

## What was changed/added
- updated `item_interaction`, removed unused variables
- added dictionary to track items and their known amount -> representation as dictionary ( key:itemspotID,value:QuantityAvailable) 
- now binding specific node references \@onready 
- update_visibility sets visibility of sprite denoting an interaction

## Testing
1. open overworld 
2. obtain an item -> should be invisible now!
3. open pause menu and return --> itemspot state prevails 
4. enter bridge scene 
5. return to overworld --> itemspot state prevails

### DIFF
--- a/Arch-enemies/overworld/entities/Player/Player.gd
+++ b/Arch-enemies/overworld/entities/Player/Player.gd
@@ -158,13 +158,9 @@ func execute_interaction():
 			enter_bridge_scene(bridge_edge)
 			
 		Interactable.InteractionType.ITEM: 
-			print("obtained item")
-			set_interactionLabel(interaction_data["text"])
 			var received_item:Item.ItemType = interaction_data["item"]
 			# adding to inventory 
 			SingletonPlayer.add_to_inventory(received_item)
-			# updating ui 
-			# adding to inventory! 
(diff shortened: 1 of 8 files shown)


mango-gremlin (approved):
works as described, thanks so much!

PR #299, opened 2024-01-21
Overworld | add Npc name to dialogue 

ScatteredDrifter:
## Motivation
Adds name of the given npc to every dialogue box they have.
Does not resolve a specific issue ( was discussed internally )

Also went over the implementation of the Dialogue-System and added / rewrote some minor portions to improve readability. 

**Warning**:
With this change names are set the following: 
1. each Dialogue contains a variable "npc_name" which denotes the name to display. Its "" at first, so empty
2. To update this dynamically we have to set this name whenever the npcs linked to the dialogue are initialized ( before that we cannot obtain their name cus its contained in their construction / object, which is only available after they've been loaded at least once ) 
3. then the name is statically taken from the internal variable set in `Dialogue_Data`

## What was changed/added
- renamed few things in `Dialogue` `Dialogue_Data` 
- added method to retrieve npc_name 
- added method to set npc_name from `npc_interaction` --> upon `_ready` is executed 
- added _Warning_ to improve this later --> setting it in npc_interaction leaves its scope / should be done elsewhere in my opinion

## Testing
Try talking to the npcs available:

image

image

image

### DIFF
--- a/Arch-enemies/overworld/dialogue/Dialogue.gd
+++ b/Arch-enemies/overworld/dialogue/Dialogue.gd
@@ -8,6 +8,7 @@ var active_dialogue:Dialogue_Data = null
 var current_npc_id : int = 0
 
 # no type here, on startup, this will be configured
+# FIXME declare type
 var npc_control_instance = null
 
 # --- /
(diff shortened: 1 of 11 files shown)


mango-gremlin (approved):
looks really good, thanks!
As I've said maybe we can make the font size bigger&put a box around the npc name to make it stand out, but as this is UI I can do that really quick.

PR #289, opened 2024-01-19
Constructed final bridge levels and overworld

mango-gremlin:
**WHEN TESTING: DELETE YOUR OLD SAVE FILE (SEE WIKI) OR THE FOX WILL LAND IN THE WATER**

## Motivation
closes #281 
closes #258
closes #257 

The final game as shown in Wiki/Story had to be implemented.

I apologize for the gigantic PR :(

### Bridges
- added the three levels: 0_Tutorial, 1_FrogHazard, 2_Final
- changed fox jump height because for some reason, fox did not jump over 3 blocks -> i thought we had said 3 blocks so a fox can clear a deer with a jump, which might be necessary when you dont have many animals to build stairs with. If this should be changed, say so, but we then need to add more animals, and it might be a bit annoying for players if they cant just jump over.
- changed fox hitbox so platforming with hazards is more satisfying -> the hitbox doesnt extend past the fox anymore, so you only die when you think you should
- pause menu location adjusted -> now always in the middle
- aspect ratio adjusted so levels do not look squished

**unfinished:**
**problems/bugs: bramble in level 1 overlaps with new inventory (doesnt really matter actually)**

### Overworld
- added the final overworld with four islands, one main one and three smaller ones (as seen in Story)
- added items drawn by @Kabooni ; added function that automatically assigns item sprite to the right item & updated items in backend to be the new items
- added all npcs that were planned, as well as functionality for the quests, which items to bring them etc
- added all bridges and connected them via bridge edges
- added the theme to the dialogue box -> UI needs to be further adjusted
- UI for inventory updated: shows sprites of animals and items next to amount you have
- UI for dialogue boxes now fits with the theme, increased font size
- last level solved -> u can now cross the bridge; @ScatteredDrifter noticed I forgot to add the tiles with no collision
- Inventory now updates and deducts animals used during solved levels

**unfinished:**
**problems/bugs: -**

**will be done in separate issues:**
- Starting Cutscene pictures  #300 
- Dialogue text, include npc portraits #303 

**probably no time (if someone has time, feel free to do this): final bridge interaction appears only when quest done**

## Testing
I'll include some screenshots when I've eaten lunch lol

## Additional Information
if anything else was forgotten, needs to be added, pls say so. First take a look in the 'unfinished' section

PR #300, opened 2024-01-21
overworld | added cutscene that plays on startup

mango-gremlin:
## Motivation
(no issue; this was initially part of #289 but the pr is so huge I tried to break it up)

## What was changed/added
Added a scene starting_cutscene.tscn, where you can flip through the three starting cutscene pictures.
This scene plays when pressing start game in the main menu;
When pressing "skip" at any time, or start game in the last picture, it sends you to the overworld.
This was the easiest way to do it; I dont know if this will fuck with anything.
Essentially I Just diverted the path from main menu scene to overworld scene to have another scene in between. Hope that works.
It also plays the music.

**pictures are not final, i just tried to do them quickly.. I hope they convey what they are supposed to, if they don't; I'll change them, but i didnt put too much work into it on purpose, I will make them nicer if there is time after the playtest.**

## Testing
https://github.com/mango-gremlin/arch-enemies/assets/104830903/cf9adbaf-dd1f-4580-bbbd-4501a75101fa


Kabooni (approved):
I love them! Sure it's a sketch, but I can see the story you want to tell.

Raffael-Rack:
Looks very nice!

PR #302, opened 2024-01-21
overworld | indicators for interacting with npcs, items, walking

mango-gremlin:
## Motivation
Only We know You Have To Press E to Interact. And WASD To Walk.

## What was changed/added
Added two sprites: An animated floating "E" Button and a WASD Sprite that disappears upon first movement.

## Testing
- wasd indicator disappears when you walk for the first time. however, if you enter another scene and come back, it will be there once again. I would leave it this way for the playtest, and i know how to change it so it saves it (probably) but i simply do not have time rn..
- item indicator disappears with item
- npc indicator floats above npcs for all eternity. Not above the deer, I'm not sure why; this will have to be fixed after the playtest...

https://github.com/mango-gremlin/arch-enemies/assets/104830903/75f38fc4-23c6-429b-bf5a-272da98fbbb5


Kabooni (approved):
lgtm

PR #310, opened 2024-01-25
Feature/disable save management

ScatteredDrifter:
## motivation:

disables save management because our game does not maintain a state over session of playing it.

## What was changed/added
1. disabled loading singleton values from save file
2. disabled saving state to file

## Testing
well test it please, you should start at the same position when starting the game + the inventory is set to default too.


aylinfi (approved):
thank you, no more manually deleting files :)

PR #312, opened 2024-01-25
overworld | allow replaying bridge level after it was done

ScatteredDrifter:
## Motivation: 

It was requested that replaying a solved level should be allowed. 

## Things changed 
removed check whether it was solved already 
-> allowing to enter a bridge-level that was done before.


aylinfi (approved):
lgtm

PR #316, opened 2024-01-26
center the dragpreview of the animals

### LINKED ISSUE #315: bridges | center the dragpreview
## Motivation
The placement of animals was based on the cursor at the lower left corner and not the dragpreview which was confusing for players. 
It would be nicer to have the cursor be in the center of the dragpreview.

## Implementation
The dragpreview is a different scene, so `_can_drop_data()` and `_drop_data()` also need to be called for THAT scene.

There also need to be adjustments because the missing offset.

aylinfi:
## Motivation
closes #315

The placement of animals was based on the cursor at the lower left corner and not the dragpreview which was confusing for players.
It is nicer to have the cursor in the center of the dragpreview.

## What was changed/added
The problem with centering the cursor on the dragpreview was that `_can_drop_data()` wasn't called anymore and nothing could be placed. This was due to the preview being a different scene, not connected to our `_can_drop_data()` function.

This could be solved by adding it and also the `_drop_data()` function to `DRAGPREVIEW.gd` which is connected to the dragpreview.tscn scene. These new functions just call the old functions in `grid_drag.gd`(this is not the greatest coding but it works).

Other than that the new positions of deer and snake needed to be adjusted, because we were using different grid positions before.

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/776f7693-a407-47bc-b3ee-dec154cfef89

### DIFF
--- a/Arch-enemies/bridges/scenes/dragpreview.tscn
+++ b/Arch-enemies/bridges/scenes/dragpreview.tscn
@@ -3,6 +3,7 @@
 [ext_resource type="Script" path="res://bridges/script/DRAGPREVIEW.gd" id="1_dtuo0"]
 
 [node name="dragpreview" type="TextureRect"]
+top_level = true
 offset_right = 40.0
 offset_bottom = 40.0
 script = ExtResource("1_dtuo0")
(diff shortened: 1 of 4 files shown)


Kabooni (approved):
Lgtm, thanks a lot!

PR #331, opened 2024-01-27
bridges | add goal indicator

### LINKED ISSUE #327: bridges | add goal indicator
## Motivation
We don't have any clue for the player to reach the other side.

## Implementation
Draw an indicator (e.g. a flag or something, but forest-themed) and place it on the other shore. Move the goal area more to the right edge, so the player is not triggering it immediately.

Kabooni:
## Motivation
closes #327
We need an indicator of where the player has to go.

## What was changed/added
1. Draw a wooden sign sprite
2. Add it to the goal scene and adjust it in the bridge levels

## Testing
grafik


mango-gremlin (approved):
looks really nice, thanks!

PR #333, opened 2024-01-28
overworld | Added custom names to dialogue system

### LINKED ISSUE #332: Add ability to set name for each dynamic dialogue page seperately
Add ability to set name for each dynamic dialogue page seperately

DerSimi:
## Motivation
This adds custom names to the dialogue system.
closes #332 

## What was changed/added
page_name() was added to the dialogue system, if no name is applied, i. e. the method not overriden, the npc name is applied. The npc name is also applied if the page name is ""

## Testing
image
image

Please insert the names into OverworldDialogues.gd
image


mango-gremlin (approved):
tested it after adding the names to the dialogues, everything works as described, thanks!

ScatteredDrifter (approved):
thank you for the contribution! 
Rand through your changes and also the addition / implementation by glenn. 

Both work well, will approve

PR #340, opened 2024-01-28
Resize one bramble and scale final level

### LINKED ISSUE #339: bridges | rework final level
## Motivation
Make the thorns in the upper right a bit smaller to allow players to build over them.

## Implementation
Change tiles.

aylinfi:
## Motivation
closes #339 

Playtesters really wanted to build above one bramble, but it wasn't possible. So make it possible!

## What was changed/added
Made the thorns in the upper right a bit smaller to allow players to build over them.
Also resized/moved final level to fit camera size better and allow for more space to build above thorns.

## Testing

fox_over_bramble


Kabooni (changes requested):
Looks great, thanks! Just one thing, the third level now starts with the pause menu visibility as true. But I'm not quite sure why, the code changes look fine.

aylinfi:
Yeah oops that was a mistake.

Kabooni (approved):
Thanks!

PR #342, opened 2024-01-28
Adjust level buttons

### LINKED ISSUE #341: bridges | adjust bridge level buttons
## Motivation
Playtesters struggled with some ui elements.

## Implementation
- put reset button in lower right corner
- make pause menu button ("esc")
- make buttons in fox mode invisible

aylinfi:
## Motivation
closes #341

Playtesters struggled with some ui elements, so we needed to adjust them.

## What was changed/added
- the reset-button has been moved to the lower right corner
- an escape-button has been added to the lower left corner that has the same function as pressing "esc"
- reset and one-step-back-button are invisible while fox-mode is on

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/116217918/005597e3-58ad-46f2-99e9-0df51b9fdd78


Kabooni (approved):
Lgtm and I didn't find any bugs

PR #345, opened 2024-01-30
overworld | adjusted collisions for bridge starts

### LINKED ISSUE #265: overworld | improve collision for bridge-starts
## Motivation

We might have to **improve** the **bridge-start** collision as they sometimes make movement around / through those bridge-parts clunky / stuck. 

## Implementation
Its primarily updating / likely decreasing those collisions to make it more pleasant

mango-gremlin:
## Motivation
closes #265 

In the playtest ppl were getting stuck on the bridge starts and I also meant to fix this for a while.

## Testing
As you can see the collision for the treestump is now only in the middle of the treestump and not as spikey, making it less clunky.

https://github.com/mango-gremlin/arch-enemies/assets/104830903/c2f761cd-21f3-4734-90fc-de767a2f713a


DerSimi (approved):
LGTM

PR #350, opened 2024-01-31
overworld | display dialogue if quest solved and not spoken to before

### LINKED ISSUE #329: overworld | indicate that quest was solved
## Motivation
We should indicate to the player earlier that the quest was solved. When the quest is solvable, add the "you give item to animal" text and add the quest-solved dialogue.

## Implementation
Add the appropriate texts
In detail:
- When speaking for the first time without the quest item, do nothing new
- When speaking for the first time with the quest item, do normal dialogue + solved dialogue together
- When the quest is done, do only solved dialogue

-> add "you give item to animal" at the beginning of quest solved dialogue

ScatteredDrifter:
## Motivation
resolves #329 

## What was changed/added
added variable to dialogue_data that tracks whether we spoke to the npc before.

Upon solving their quest before speaking to them the dialogue will be entered again ( and display the solved one ). 
It will then set this variable  to true preventing this to loop forever. 
This adds an invariant. 

## Testing
Well test the following: 
1. gather nut and fly
2. speak to squirrel and observe behavior 
3. speak to spider and observe behavior 
4. speak to snake afterwards

### DIFF
--- a/Arch-enemies/overworld/dialogue/Dialogue.gd
+++ b/Arch-enemies/overworld/dialogue/Dialogue.gd
@@ -36,12 +36,20 @@ func in_dialogue() -> bool:
 	return is_in_dialogue
 
 func exit_dialogue():
-	is_in_dialogue = false 
-	active_dialogue = null
-	print("exiting dialogue")
+	# quest could be removed if it was solved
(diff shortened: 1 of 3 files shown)


ScatteredDrifter:
now actively tracked in #352

PR #351, opened 2024-01-31
Better sounds, sound settings

### LINKED ISSUE #295: (Optional) Improve Sounds
## Motivation
Some stuff is still open from #280 and the associated PR. If we have time and/or suffient feedback from playtesting, we can implement (some of) them as required.

## Tasklist
- [ ] turn down background volume while victory music is playing
- [x] improve dialogue signal (use singleton instead), currently it takes a cursed detour over the world scene
- [ ] animal specific dialogue
- [x] snake sound to harsh
- [ ] improve step sound
- [x] maybe background music as autoload, makes it easy to manipulate volume through code and enables smooth transitions between scenes
- [x] **sound settings**
- [x] **button sound**
    - [x] bridges: pause menu, settings menu, drag, step back, reset, tooltip
    - [x] overworld: pause menu, settings menu, 
    - [x] main menu
    - [x] cutscene
    - [x] about page

marcel-iv:
## Motivation
closes #295 

## Changes
- a sound settings menu now exists: you can access it from the pause menu, either from the overworld, the levels, or the main menu. Note that there's actually 3 different sound menus (main menu, overworld, bridges), this was due to the existing implementation. A more efficient solution would have required a major overhaul of existing scenes, which is way outside the scope of this issue. 
   - you can change main volume, music volume and effects volume, and this is carried over to other scenes, and is only changed when pressing the appropriate button.  
- the sound for the snake was improved to be a little bit less harsh
- all button elements now have a sound attached to them. However, since some buttons cause a change to a different scene, this causes some sounds to be cut off. Moving an AudioStreamPlayer to an Autoload node would be the best solution, but that would be cumbersome to change. If thats a large problem, I can *try* to fix it on friday.
- some minor refactoring was done, such as using a single script for effects, instead of two. 
- a small bug was fixed, now walking sound isn't looped when switching to drag mode while walking.
- background music is now a global node, which means its seemlessly looped even when transitioning to another scene. 
- there's a globally accessible setting script, which right now is only used to control volume, but might be interesting in future implementations. We decided against abusing the singleton further.

## What hasn't changed
- @aylinfi suggested reducing the volume of the background music when victory music is playing. However, this would take some more work done (using a Timer Node), and since the sound is barely over a second long and doesn't loop well, I decided against implementing it.
- The walking sound remains the same, although there are some alternatives in the asset folder if you're curious. However, we decided to remain with the current sound.

## Known Issues
- in the final level, the settings menu is larger than the pause menu, which looks a bit weird, this is due to the existing implementation as an overlay, and according to @koennixer, extremely annoying to change

## Testing
Button sounds work, and the level still works just as before.

https://github.com/mango-gremlin/arch-enemies/assets/104799849/4469b8bf-3b96-4afd-b711-1eeeb078844d

Here you can see (and hear) the menu in a bridge level and the overworld

https://github.com/mango-gremlin/arch-enemies/assets/104799849/81444913-4e46-4feb-a680-a89980546a32

Anything else you can see on the branch yourself. We definitely recommend testing there, sound is subjective and hard to gauge with 20s videos.


Arch-enemies/bridges/script/grid.gd
```diff
@@ -450,22 +450,29 @@ func reset_modes():
...
+			pause_menu.visible = false
+			settings_menu.visible = false
+			menu_mode = false
+			change_ui_visibility()
+		# if currently in the settings menu -> return to pause menu
+		elif settings_menu.visible:
+			pause_menu.visible = true
+			settings_menu.visible = false
 
-func open_pause_menu():
```
Kabooni, line 459: The function was connected to the `esc` button and needs to be reconnected

Kabooni (changes requested):
Some things that are bothering me:
- When going main menu > Options >Return You can only go to the overworld and will skip the intro cutscene
- The volume change for the buttons in the settings menu is only applied when leaving the menu and going back

Might be more, but I'm too tired for it now.
Overall, thanks a lot, it's a lot of work you put into this! And in general it works really well

Arch-enemies/bridges/script/grid.gd
marcel-iv: This implementation is based on an older implemenation on how the pause menu works, and I only merged main into the branch later, thats why I deleted it. Its probably possible to re-implement it, but since this one works as well (even though its much uglier), I'll keep it. If someone else wants to do that, be my guest, but I don't have the time for refactoring

marcel-iv:
> * When going main menu > Options >Return You can only go to the overworld and will skip the intro cutscene
> * The volume change for the buttons in the settings menu is only applied when leaving the menu and going back

Good catch, *should* both be fixed now

Arch-enemies/bridges/script/grid.gd
Kabooni: I mean we can't just leave it as it is. Right now the esc button does nothing, so we should either remove it or find a way to connect it to the pause menu in the new implementation, which I would prefer. I will look into it this evening, maybe I can do it

Kabooni (approved):
I've reconnected the esc button and with your new changes, you could merge it.

mango-gremlin (commented):
tested it as well, very nice!
the dialogue buttons dont make a sound but i assume this was intentional as constant button sounds in dialogue might be annoying.

PR #352, opened 2024-01-31
overworld | improve quest dialogues

### LINKED ISSUE #329: overworld | indicate that quest was solved

Kabooni:
Something went wrong with the branch and I (Kabooni) couldn't commit to it. Here's a new updated pr.

## Motivation
resolves #329 

## What was changed/added
added variable to dialogue_data that tracks whether we spoke to the npc before.

Upon solving their quest before speaking to them the dialogue will be entered again ( and display the solved one ). 
It will then set this variable  to true preventing this to loop forever. 
This adds an invariant. 

Also added new dialogue to inform the player that an item was given to the npc.

## Testing
Well test the following: 
1. gather nut and fly
2. speak to squirrel and observe behavior 
3. speak to spider and observe behavior 
4. speak to snake afterwards

The quest solved dialogue now contains an action dialogue for giving an item. E.g. for the spider:
image

### DIFF
--- a/Arch-enemies/overworld/dialogue/Dialogue.gd
+++ b/Arch-enemies/overworld/dialogue/Dialogue.gd
@@ -36,12 +36,20 @@ func in_dialogue() -> bool:
 	return is_in_dialogue
 
 func exit_dialogue():
-	is_in_dialogue = false 
-	active_dialogue = null
-	print("exiting dialogue")
+	# quest could be removed if it was solved
 	if SingletonPlayer.check_dialogue_finished(current_npc_id):
 		SingletonPlayer.remove_quest_string(current_npc_id)
-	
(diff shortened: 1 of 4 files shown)


mango-gremlin (approved):
tested it, works nicely
Github desktop did somethig with this branch: I cannot find it with github desktop, but with git bash it works...

PR #361, opened 2024-02-01
overworld | dialogue playing twice if quest solved

ScatteredDrifter:
## Motivation
resolves a bug that was added after merging #352 :

Specifically one could talk to a npc, get their requirement and submit the quest. Once that was done the dialogue for solving the quest was shown twice - because the internal tracking of "spoken_to_before" was not updated on each occasion. 

## What was changed/added
To resolve the internal state is now updated upon exiting the dialogue - every time. 

## Testing
As in #352 you can try the following:
**solve quest without prior talk**: 
1. gather item for one npc 
2. talk to npc and observe whether the quest done dialogue is shown after the initial dialogue

**solve quest with prior talking**: 

1. talk to npc 
2. gather their requirement 
3. talk to them again and observe the quest_done_dialogue to show up only once! 
---


Kabooni (approved):
I've tested both ways and it works as intended now, thanks!

PR #363, opened 2024-02-02
bridges | Made the starting branch more visible

### LINKED ISSUE #356: bridges | starting tree branch more clearly
## Motivation

The playtesters were wondering why the fox is flying in the air, because the branch wasn't seen.

Raffael-Rack:
## Motivation
closes #356

Some playtesters were wondering why the fox starts out floating.

## What was changed/added

Changed the left segment of the background to emphasize the branch. The brighter green contrasts with the dark brown of the tree and the black outline of the fox and should make the branch more apparent.

## Testing

Here's the new look
Screenshot Issue 356


marcel-iv (approved):
Looks good, and communicates the purpose of the starting point very well in my opinion, although the bright green doesn't look too much like a branch to me.

I also noticed that the tutorial level doesn't use the branch as a starting point, which is a bit odd and maybe adds to player confusion. I don't think this is anything worth changing in the last few hours before a deadline though, so feel free to merge.

Kabooni (approved):
I think the purpose is now clearer than before, but the branch might still be overlooked. But honestly, I think to make it 100% clear for the player we would have to change the approach (and e.g. make like a little edge with the tileset), which is not necessary for the presentation.

mango-gremlin (approved):
nothing was broken by the removal, idk what that was about. investigation ongoing, @Kabooni will merge when it was determined whats going on

Kabooni:
I've re-added the missing node in the scene file, which was tedious because godot wanted to delete it all the time, but now it's there again. In general there was no difference in gameplay with or without it

PR #366, opened 2024-02-02
overworld | collision size and pos of npc interaction determined from sprite

### LINKED ISSUE #365: overworld | collisions for NPCs
## Motivation
You can walk through npcs

## Implementation
Add a collision box at least around parts of the npc!

mango-gremlin:
## Motivation
closes #365 

NPCs had no collision as of the first playtest.

## What was changed/added
Added a collision shape (circle), depending on the npc sprite, the position and size of it is changed to approximate the sprite.
Also added collision to the frogs on the frog island (these are just the frog hazard tile)

## Testing

https://github.com/mango-gremlin/arch-enemies/assets/104830903/0a295b68-9451-4411-aaac-d06136e501bd

### DIFF
--- a/Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
+++ b/Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
@@ -144,20 +144,29 @@ func obtain_required_item() -> Item.ItemType:
 func set_npc_sprite():
 	# initialize reference
 	var referenced_sprite:Sprite2D = $Sprite2D
+	# collision shape for size change
+	var npc_collision = $StaticBody2D/CollisionShape2D
 	
 	if npc_special_sprite != null:
 		referenced_sprite.texture = npc_special_sprite
+		npc_collision.position = Vector2(0,0)
+		npc_collision.scale = Vector2(0.5,0.5)
 		return
 	var path_to_sprite:String 
 	match npc_animal_type:
 		Animal.AnimalType.SNAKE: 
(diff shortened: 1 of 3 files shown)


Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
```diff
@@ -144,20 +144,29 @@ func obtain_required_item() -> Item.ItemType:
...
 	if npc_special_sprite != null:
 		referenced_sprite.texture = npc_special_sprite
+		npc_collision.position = Vector2(0,0)
+		npc_collision.scale = Vector2(0.5,0.5)
 		return
 	var path_to_sprite:String 
 	match npc_animal_type:
 		Animal.AnimalType.SNAKE: 
 			path_to_sprite = "res://assets/art/characters/snek.png"
+			npc_collision.position = Vector2(0,-1)
```
ScatteredDrifter, line 159: Reasonable addition. I would argue that it should be moved to another position at a later point ( like only calling "create Npc sprite" which then loads the sprite and collision. 
However the current structure seems reasonable for now too

ScatteredDrifter (approved):
Looking good overall, i amended some consideration regarding the code structure. 

I cannot test locally right now however given with your video it should operate accordingly

Arch-enemies/overworld/entities/interaction_overworld/npc_interaction.gd
mango-gremlin: yes that is fair, this could be refactored, I will see if i can do that on the weekend.

Issue #307, opened 2024-01-23
Tasklist post-playtest


marcel-iv:
## Motivation
This is a collection of everything that we want to implement post-playtest, either as a request by the playtesters, or other things we noticed along the way. Expand this list with everything we mentioned in old PRs as future issues.

## Things that will not be fixed for v.1.0.0
- Dialogue formatting (bold/italic)
- #346
- more recognizable squirrel / spider / conspiracy board
- (low prio) end scene
- #369
- flip animals
- schlange sieht so aus wie wenn man sie neben spider mit kopf placen könnte (?)
- Overworld: prioritize movement by last pressed button

# Things that were fixed

## UI
- [x] #354 
- [x] #325

## Dialogue
- [x] #349
- [x] add name of speaking animal, not quest animal
- [x] "previous" button should be greyed out, don't delete
- [x] #329
- [x] #336
- [x] #326
- [x] main quest: get to the beavers

## Art
- [x] ~more flies in the fly item~ fixed by changing dialogue
- [x] #335 
- [x] cutscene more recognizable

## Overworld
- [x] #347
- [x] Item list doesn't update all the time
- [x] #367
- [x] #365
- [x] dialogue indicators should only appear for dialogue
- [x] #362
- [x] remove save file
- [x] #332 
- [x] fix deer NPC

## Bridges
- [x] move reset to bottom right corner
- [x] #335
- [x] #327
- [x] make reset/stepback invis during fox mode
- [x] #356
- [x] #313
- [x] #343
- [x] #339
- [x] #359 
- [x] #311 
- [x] fill edges

## Tutorial
- [x] #357

## Sounds
- [x] #295 

## Movement
- [x] #328

## Other
- [x] #374
- [x] #373 and #379

ScatteredDrifter:
For **overworld**
- [ ] remove saving coordinates and items **or** implement complete saving system 
- [ ] simplify creating dialogues
- [ ] allow dialogues without any quest attached
- [ ] restructure dialogue system --> should not require active_dialogue as reference point to **obtain state of dialogue-entries** --> this should be communicated in the module itself
- [ ] refactor **singleton** --> remove the npc-creation methods! --> don't belong there --> those should be contained in a class that can be called to **initialize all** dialogues and return the dictionary which is currently created in the singleton

mango-gremlin:
this was all fixed in the current release. Everything that wasnt fixed is tracked in #380

PR #370, opened 2024-02-02
added another pic to show frog is evil, cleaned up old pics

mango-gremlin:
## Motivation
in #307 -> one point was making starting cutscene clearer.

## What was changed/added
added another pic to show frog poisoning beaver, and beaver to the dam so its recognisable as dam.
cleaned up the other pics.

## Testing
cutscene_1_v2_2
cutscene_1_aftermath


Kabooni (approved):
Works and looks awesome, I love it!

PR #376, opened 2024-02-04
Added Final export for web and windows

mango-gremlin:
## What was changed
- I added an exports folder containing both the windows (.exe) export and the web export.
- uploaded the final game to itch: https://mango-gremlin.itch.io/arch-enemies
- the page is a bit empty; will add some screenshots momentarily

## Testing
**It would be nice if some of you could test the game on itch and the .exe version to make sure everything works**!!!!!


marcel-iv (changes requested):
Stuff I've found: 
Both version: 
Snake Portrait for dialogue went missing (again), probably a naming thing (again)

Web Version only:
Some symbols aren't rendered correctly in deer and final dialogue
image
image

Also if its possible, it would be cool if our Toadally Anonymous sprite could be the sprite used on the browser tab and icon for the executable. Only if you have time tho. 
image

image

Kabooni (changes requested):
For the itch version:
- The about page was formatted weird:
grafik
- The settings page, bridge pause menu and goal menu got blurry
- Bridge inventory got blurry:

Blurrying is something that might not be fixable, but at least the about page should be adjusted

marcel-iv:
Interestingly, the blurry menus did not appear on my end, maybe thats browser/screen resolution specific? (Chrome, 1920x1080 in my case).

About page was still the same as in the screenshot by @Kabooni though

Kabooni:
>Interestingly, the blurry menus did not appear on my end, maybe thats browser/screen resolution specific? (Chrome, 1920x1080 in my case).

Hm, I have the same resolution and was using firefox, and got the same results in edge.

Raining-Cloud:
> > Interestingly, the blurry menus did not appear on my end, maybe thats browser/screen resolution specific? (Chrome, 1920x1080 in my case).
> 
> Hm, I have the same resolution and was using firefox, and got the same results in edge and chrome. And it's just the font that's blurry in those scenes, the rest is fine. So I'm really confused what the problem is

Like this ?
grafik

Kabooni:
> Like this ? grafik

Yes. Or another example
grafik

marcel-iv:
Ah I just realized I didn't play the game in fullscreen mode, where the blurring isn't really noticable. So this occurs on every system, my bad.

mango-gremlin:
> Stuff I've found: Both version: Snake Portrait for dialogue went missing (again), probably a naming thing (again)

Yep, idk why github sneaks that in, will change the file name back...

> Web Version only: Some symbols (only `(`and `)` I think) aren't rendered correctly in deer and final dialogue 

That's so weird. I dont really know where that comes from and how to fix it. If it's only those symbols, I will just replace them i guess.

> Also if its possible, it would be cool if our Toadally Anonymous sprite could be the sprite used on the browser tab and icon for the executable. Only if you have time tho. 

For web: I did think about that, however that icon on itch is also the one that it shows you when the game is listed somewhere on itch like so:
image
And i thought for that it might be nicer if we had the thumbnail. Thats easy to change to the frog tho, if you want.

For Windows: the thing is i tried to do that, but it did not work lol. Will look up how to do this and try again.

> About page fucked on web (and windows ?)

Dang I have no idea how that happened... will see what I can do there. Might have to do this this evening.

> Blurred text

@Kabooni this is something we talked about, because its also a problem in the debug mode in godot; but i dont know if there is a solution to this? :( It might just be something we have to deal with, feel free to also try to look for a solution online @Kabooni @marcel-iv @Raining-Cloud , because i dont know where this is coming from.

Raining-Cloud:
> Ah I just realized I didn't play the game in fullscreen mode, where the blurring isn't really noticable. So this occurs on every system, my bad.
I think this is related to the font-rendering, it also occurs in the godot project for me. Never thought its an actual Problem

Kabooni (approved):
To make it clear - my problem is not in the fullscreen version, but without it. **But:**
I've found the problem, it's the zoom of my browser. For some reason the rest of game is scaling correctly, but not the font lol

After the about page on itch is updated as well, it could be merged

mango-gremlin:
I have now fixed all mentioned issues (except the font size thing), exported both versions and uploaded the new web game to itch, so you can test it there.

The font thing:
- its because pixel fonts only work at specific font sizes without being blurry, which is sad
- we would have to adjust every font size in the game, which is a bit tedious 

-> the font problem can be fixed with the browser zoom as @Kabooni said

marcel-iv (approved):
Everything looks good, and works great on both versions, except that which was already mentioned.

Also the itch page looks really great, thanks for that!

PR #379, opened 2024-02-06
added trailer to readme

mango-gremlin:
## Motivation
Trailer would be nice

## What was changed/added
Added trailer to readme


koennixer (approved):
This is fine. The railer is well placed and runs.

PR #382, opened 2024-02-11
removed old bridge files, renamed and restructured the rest

### LINKED ISSUE #236: bridges | remove outdated scripts and scenes
## Motivation
Currently our repo is a bit clustered with outdated code from the previous implementation. For the sake of clarity, we should remove that. 

## Implementation
Remove all scripts/scenes not currently in use, and check if nothing breaks (including stuff on the overworld).

marcel-iv:
## Motivation
closes #236 

## What was changed/added
- removed every script/scene that was part of the previous bridges implementations
- added an actual folder structure to bridges, to more easily find scripts/scenes
- moved sounds out of their respective folders, into the common sound folder
- moved assets out of the seperate bridges/assets into the common assets folder
- changed parameters to be optional in case they aren't used, to cut down on the amount of warnings for easier debugging
- renamed existing filed to conform to snake_case, changed relevant scripts accordingly
- moved a collision shape to prevented falling out of the frog level

## What isn't changed
- folder structure in assets and overworld remains largely the same
- naming still doesn't follow conventions in assets/overworld
- overworld/assets likely contains old files

## Testing
Everything works the same as before, which is hard to show in a single video, so I showed just a simple video of doing everything in one level. If possible, check everything on the branch for yourself, in case I missed some obscure reference to a deleted/renamed file.

https://github.com/mango-gremlin/arch-enemies/assets/104799849/a6b0d55e-5572-4212-b533-e5467fade089


Kabooni (changes requested):
I've got an error for the button sounds
grafik
Which I think is only happening in the pause menu "options" button, and return button in the sound menu.
Also the template/old level has some tilemap problems
grafik

Besides that, I didn't find any problems.

marcel-iv:
The error should be fixed now, it was caused because the sound was set to be played after transitioning to a new scene.

The template scene (formerly Grid.tscn) looked like this ever since the hazard tileset was updated, its probably better to fix this in another issue.

mango-gremlin:
Do we need the old/template bridge level? Like now that we have the actual bridge levels, we can delete it, right? especially if it isn't even working correctly. It isnt used for anything, and doesnt contain anything that's not in any of the other bridge levels (so if you wanted to make a new bridge level, you could just copy the tutorial level)
It was necessary as a place to test new features but now that levels are implemented and it has some kind of issue with the sprites I don't know whether it's worth keeping

marcel-iv:
Technically we can delete the template, but I'm in favour of cleaning it up and preserving it. Right now, to create a level, you duplicate the template level. You could use one of the existing levels ahead, but using a dedicated template is better in my opinion. 

I could clean it up as part of #369, since that (probably from what I've) also involves some cleanup on existing levels.

Kabooni (approved):
I still have the problem that the options button in general and the return button of the settings menu have cut-off sounds to the point where I sometimes don't hear them at all. But the error has been fixed and the problem is not really related to the pr here, so we could think about a solution in a different issue.

Kabooni:
> Technically we can delete the template, but I'm in favour of cleaning it up and preserving it. Right now, to create a level, you duplicate the template level. You could use one of the existing levels instead, but using a dedicated template is better in my opinion.

I agree, it's always good to have a separate template, as long as it's very basic and easily adjustable, which our is as far as I know.

